PR #32, opened 2026-02-16
feat: add MDS030 empty section body rule (plan 63)

jeduden:
## Summary
- add new MDS030 / empty-section-body rule to report headings with no meaningful section body
- add configurable scope (min-level, max-level) and explicit exemption marker (allow-marker)
- add rule docs and fixtures under internal/rules/MDS030-empty-section-body
- integrate rule imports in CLI/config/integration test entrypoints
- update existing fixtures/proto templates and plan docs to account for the new rule semantics

## Validation
- `go test ./...`
- `go tool golangci-lint run`
- `go run ./cmd/mdsmith check .`

## Plan
- execute plan 63 (plan/63_empty-section-body-rule.md) and mark it complete

### DIFF
--- a/internal/rules/MDS030-empty-section-body/README.md
+++ b/internal/rules/MDS030-empty-section-body/README.md
@@ -111,7 +111,8 @@ This section explains what the command does.
 
 ## Edge Cases
 
-- Parent headings are considered non-empty when nested subsections contain
-  meaningful content.
+- Parent headings are considered non-empty if any meaningful content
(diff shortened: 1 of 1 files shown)


internal/rules/MDS002-heading-style/bad/default.md
```diff
@@ -8,3 +8,5 @@ diagnostics:
 
 Section
 -------
+
```
jeduden: bad shouldnt be fixed.

PR #183, opened 2026-04-27
Add pre-merge-commit hook to resolve file ordering race

jeduden:
## Summary

This PR adds a `pre-merge-commit` git hook to the merge driver installation process. The hook re-runs `mdsmith fix` after git has resolved all per-file merges, fixing a race condition where generated sections could become stale when multiple source files are modified in a merge.

## Problem

When merging branches that both modify a generated file (e.g., `PLAN.md`) and its source files (e.g., `plan/*.md`), the per-file merge driver runs on `PLAN.md` before git has finished merging the source files. This causes the regenerated catalog to reflect an incomplete view of the sources, resulting in stale generated content.

## Solution

- **New `ensurePreMergeCommitHook()` function**: Creates `.git/hooks/pre-merge-commit` that runs `mdsmith fix` on all registered files after git completes all per-file merges
- **Hook management**: The hook includes a marker (`preMergeCommitHookMarker`) so `merge-driver install` can safely replace managed hooks without clobbering user-authored hooks
- **Resilient file handling**: The hook checks for file existence before fixing, allowing it to survive schema changes and file renames
- **Installation integration**: `runMergeDriverInstall()` now calls `ensurePreMergeCommitHook()` and reports the hook path in its output

## Key Changes

- Added `ensurePreMergeCommitHook()` to generate and install the pre-merge-commit hook with proper error handling
- Added `shellQuote()` helper for safe shell escaping of file paths in the hook script
- Updated `runMergeDriverInstall()` to install the hook alongside the merge driver config and `.gitattributes`
- Added comprehensive unit tests covering hook creation, replacement of managed hooks, refusal to overwrite user hooks, and binary resolution failures
- Added end-to-end test `TestE2E_MergeDriver_FileOrderingRace_Resolved` that reproduces the original CI failure and verifies the fix
- Updated CI workflow to build mdsmith from source rather than using a pinned release, ensuring the merge driver and hook always match the directive schema of the branch being merged

## Implementation Details

- Hook content is generated with proper shell quoting to handle file paths safely
- The hook uses `set -e` to fail fast if any `mdsmith fix` invocation fails
- Existing hooks are checked for the management marker before being replaced
- The hook gracefully skips files that no longer exist (`if [ -e file ]`)
- Binary resolution uses the same mechanism as the merge driver itself for consistency

https://claude.ai/code/session_012XG9t645SL8z2RW9ik79gw

### DIFF
--- a/.github/workflows/merge-queue.yml
+++ b/.github/workflows/merge-queue.yml
@@ -25,18 +25,15 @@ jobs:
           fetch-depth: 0
           token: ${{ secrets.MERGE_QUEUE_TOKEN }}
 
-      - uses: actions/setup-go@7a3fe6cf4cb3a834922a1244abfce67bcef6a0c5 # v6.2.0
-        with:
-          go-version-file: go.mod
-
       - name: Install mdsmith merge driver
-        # Build from the checked-out source so the merge driver and
-        # its pre-merge-commit hook always match the directive schema
-        # of the branch being queued. A pinned release binary lags
-        # behind whenever the catalog/include schema evolves and can
-        # silently produce stale generated sections.
+        env:
+          MDSMITH_VERSION: v0.5.0
+          MDSMITH_SHA256: 87519781aa7b5ab147d5ab1d75d4e0a1c6213479110055972a21025b537ce171
         run: |
-          go build -o "$RUNNER_TEMP/mdsmith" ./cmd/mdsmith
+          curl -fsSL "https://github.com/jeduden/mdsmith/releases/download/${MDSMITH_VERSION}/mdsmith-linux-amd64" \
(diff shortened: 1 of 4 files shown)


.github/workflows/merge-queue.yml
```diff
@@ -25,15 +25,18 @@ jobs:
           fetch-depth: 0
           token: ${{ secrets.MERGE_QUEUE_TOKEN }}
 
+      - uses: actions/setup-go@7a3fe6cf4cb3a834922a1244abfce67bcef6a0c5 # v6.2.0
```
jeduden: we need to download releases and not build the merge driver. 
lets split the pr in to. one to fix. and later a second one to pull in the next release of mdsmith.

PR #266, opened 2026-05-12
Add security policy and supply-chain hardening documentation

jeduden:
## Summary

Establish a security policy for mdsmith and harden the release pipeline against supply-chain attacks (TanStack / mini-shai-hulud npm worm class).

## Key Changes

**New documentation:**
- `SECURITY.md` — public security policy, links to the canonical release pipeline doc and an inline catalog of security audit notes
- `docs/development/release.md` — canonical release pipeline doc: trigger, OIDC trusted publishers, `release` GitHub environment, supply-chain controls, operational checklist
- `docs/development/release-channels/` — one file per publishing channel (npm, PyPI, VS Code Marketplace, Open VSX, GitHub Releases). `release.md` enumerates them via `<?catalog?>`. `npm.md` is also the canonical home of the `@mdsmith/*` package list, with a Go test gating drift against `npmPlatformBuilds`
- `docs/development/secret-rotations.md` — rotation procedure + `<?catalog?>` over per-secret files
- `docs/development/secret-rotations/` — one file per tracked long-lived secret (currently VSCE_PAT, OVSX_PAT, MERGE_QUEUE_TOKEN), structured front matter
- `docs/development/release-tooling.md` — architecture doc: every workflow needing runtime logic calls `mdsmith-release` subcommands rather than inline scripts
- `docs/security/2026-05-12-supply-chain-hardening.md` — point-in-time threat model against the TanStack chain

**Release pipeline hardening (`.github/workflows/release.yml`):**
- `concurrency: { group: release, cancel-in-progress: false }` (tag-agnostic) serializes runs
- `environment: release` on every job holding a publishing credential (vscode, npm, pypi, release) — gates the OIDC token's environment claim behind required-reviewer rules
- `if: github.repository == 'jeduden/mdsmith'` guards on every publishing job
- `bun install --frozen-lockfile --ignore-scripts` everywhere

**CI hardening (`.github/workflows/ci.yml`):**
- `npm-lifecycle-guard` job rejects any PR adding an install-, uninstall-, prepare-, pack-, or publish-time lifecycle hook (15 hooks total) to published manifests via a jq set-membership test (`select(.key as $k | $banned | index($k))`); portable across jq 1.5+. `set -euo pipefail` so malformed JSON fails loud
- Explicit `permissions: { contents: read }` at the job level

**Automated secret rotation (Go, baked into `mdsmith-release`):**
- `.github/workflows/secret-rotation-reminder.yml` — monthly cron that opens a labelled GitHub issue 30 days before any tracked secret expires
- `.github/workflows/record-secret-rotation.yml` — `workflow_dispatch` entry point that opens a PR updating the `lastRotated` date in the matching per-secret file; PR body includes `Closes #N` so the merge auto-closes the reminder. Also runs `mdsmith fix docs/development/secret-rotations.md` to refresh the catalog body that interpolates `lastRotated`
- `internal/release/secretrotations.go` — port of the logic to Go (round-trip ISO-8601 validation, periodDays > 0, duplicate-title detection, idempotent label + issue creation). The workflow invokes it via two new `mdsmith-release` subcommands: `check-secret-rotations` and `record-rotation <title> <date>`. Table-driven tests in `internal/release/secretrotations_test.go` use `t.TempDir()` fixtures and a shell stub for `gh` so no live API calls run in tests
- Both workflows use `actions/setup-go` + `go run ./cmd/mdsmith-release …` — no separate runtime, no per-script lockfile

**Out-of-scope LSP refactor (bundled on this branch):**
- `internal/lsp/server.go` — extracted the AfterFunc callback in `scheduleLint` into a `runLintIfCurrent` method backed by a `pendingLint` identity token (allocated before AfterFunc starts the goroutine, so the callback never races on the timer assignment). `prev.timer.Stop()` and `handleDidClose`'s Stop both run outside `pendingMu` to avoid serializing concurrent scheduleLint calls. Unit tests cover the live=true, live=false, and shutdown branches

## Notable implementation details
- OIDC trusted publisher claims constrain on `environment=release` in addition to `repo` and `workflow`
- Per-file catalog patterns mean adding a new release channel, a new tracked secret, or a new release-tooling subcommand is one new file — no counts to update
- The `scripts` CI job (and its codecov flag) was earlier introduced for the TypeScript scripts and has been removed: the Go port runs in the existing `test` job under the `go` flag
- The `scripts-coverage-upload` split that isolated id-token between fork-test and same-repo-upload jobs is also gone with the TypeScript scripts; the Go tests have no per-job OIDC requirement

https://claude.ai/code/session_01NN6Xmj2CKBBnnvcX9J6nmM


jeduden:
Responding to the latest review batch in one comment (the per-thread reply API isn't reachable for the most recent threads).

**Addressed**
- `internal/lsp/server.go` (scheduleLint nil-guard comment, [20cee8a](https://github.com/jeduden/mdsmith/commit/20cee8a)): rewrote per Copilot's suggestion. The new wording states the actual guarantee — atomic allocation/timer-assignment/registration under `pendingMu` — and drops the overspecified claim about a callback-accepting constructor reintroducing the race.
- `.github/workflows/ci.yml:146` (npm-lifecycle-guard jq, [e345f06](https://github.com/jeduden/mdsmith/commit/e345f06)): the `$banned | index(.key)` form I first pushed was a real bug — `.key` was evaluated against `$banned` (the array), not the entry record, and jq errored with `Cannot index array with string "key"`. Fixed by binding `.key as $k` before the pipe.

**Not changing (false positives)**

1. **"`index/1` returns 0 for the first element, which is falsy in jq conditionals"** — `0` is truthy in jq. From the jq manual: *"`null` and `false` are considered "false values", and anything else is a "true value"."* Empirical check:
   ```
   $ echo 'null' | jq -r '0 | if . then "0=truthy" else "0=falsy" end'
   0=truthy
   $ echo '{"scripts": {"preinstall": "evil"}}' | jq -r --argjson banned '["preinstall","install"]' \
       '.scripts // {} | to_entries | .[]? | select(.key as $k | $banned | index($k)) | "\(.key)=\(.value)"'
   preinstall=evil
   ```
   `preinstall` (banned index 0) is correctly flagged.

2. **"GitHub Actions permissions do not support an explicit `none` value"** — `none` is documented and supported per-scope. From the [GitHub docs](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token): *"You can define the access that the `GITHUB_TOKEN` will permit by specifying `read`, `write`, or `none` as the value of the available permissions."* Keeping `issues: none` at the workflow level so the intent ("any future second job inherits the minimum and must opt-in to `issues: write` per-job") is structural, not a comment that can rot.

---

PR #456, opened 2026-06-02
Fix merge-queue loop: resync .gitattributes and add `merge-driver ci-install`

jeduden:
## What was looping

The **Merge Queue** workflow ran 30+ times in ~3 minutes across 4 PRs, each ending in `cancelled`/`failure`. Every run hit the same deterministic failure:

```
##[error]merging PR #445: git merge failed (exit 2): error: Your local
changes to the following files would be overwritten by merge: .gitattributes
Please commit your changes or stash them before you merge. Aborting
```

## Root cause

`merge-queue.yml` runs `mdsmith merge-driver install` before the action merges PRs into a batch branch. That command re-renders the tracked `.gitattributes` managed block from `.mdsmith.yml`'s `ignore:` list. The committed block was **stale** — four `ignore:` patterns were added to `.mdsmith.yml` without regenerating `.gitattributes`:

- `.git/**`
- `internal/rules/*/pattern/**`
- `editors/**/node_modules/**`
- `editors/**/dist/**`

So `install` left `.gitattributes` modified in the worktree, the action's local `git merge` aborted ("local changes would be overwritten"), the action **requeued** the PRs (re-adding the `queue` label), and the `pull_request: labeled` trigger fired the workflow again → **infinite loop**. The same drift made `mdsmith check .` red on `main` (MDS048 git-hook-sync), so `main` itself was already broken; every PR inherited the stale file.

## Two-part fix

This PR stops the loop **today** and makes it **structurally unreachable** going forward.

### 1. Resync `.gitattributes` (immediate fix)

Regenerate the managed block so it matches `.mdsmith.yml`. `mdsmith check .` goes green and the queue's `merge-driver install` becomes a worktree no-op (idempotent), so the merge can no longer abort on it.

A red/green dogfood guard (`internal/integration/gitattributes_sync_test.go`) asserts the committed `.gitattributes` stays in sync with `.mdsmith.yml` at the `go test` gate, mirroring MDS048's own `GlobsEqual` check. A future `ignore:` edit that forgets to regenerate `.gitattributes` now fails CI here too, so the drift can't reach `main` and bounce the queue again.

There is **no version skew**: the pinned baseline (`v0.29.0`, used by the queue) and HEAD render the identical managed block from the same `.mdsmith.yml`, so resync alone is sufficient. An earlier `merge-driver install --no-gitattributes` flag and a `git checkout -- .gitattributes` workflow restore were dropped because they defended against a skew that doesn't exist.

### 2. `merge-driver ci-install` (durable structural fix)

Add a new subcommand — the `npm ci` analogue of `install`'s `npm install`. It treats the committed `.gitattributes` as the source of truth: it **verifies** the managed block against `.mdsmith.yml` and exits non-zero on a missing file, missing block, or drift, **without ever writing** the tracked file. Verification runs **first**; only on success does it register the merge driver and install the pre-merge-commit hook (both git-internal and untracked), so a failing run leaves git config and hooks untouched. Reads are lstat-guarded against symlinks, like `install`.

This makes the loop unreachable by construction: the queue can't dirty the worktree, so the merge always starts clean, and real drift fails one setup step (which doesn't requeue) instead of looping. The queue can adopt it once the pinned baseline ships it — `merge-queue.yml` carries a `NOTE` to switch the install step to `ci-install` then.

## Scope

- `.gitattributes` resync + `internal/integration/gitattributes_sync_test.go` dogfood guard
- `cmd/mdsmith/mergedriver.go`: `ci-install` subcommand + `verifyGitattributes` (symlink-guarded, never writes)
- unit + e2e tests for `ci-install`: in-sync (no write), missing file, no managed block, drift, glob-arg rejection, and the config / register-driver / unmanaged-hook / read / non-regular-file error paths
- `docs/reference/cli/merge-driver.md`: `ci-install` section
- `.github/workflows/merge-queue.yml`: `NOTE` to adopt `ci-install` after the next pinned-baseline bump

## Verification

- `mdsmith check .` → `exit=0` (391 files, 0 failures); was `exit=1` (MDS048) on the stale file
- `go build ./...`, `go vet`, `go test ./...`, and `golangci-lint` all pass
- 100% patch coverage on the new `ci-install` code

---
_Generated by [Claude Code](https://claude.ai/code/session_01XHacbtgTVycBpi9XBX9bkm)_

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -4,10 +4,14 @@
 # BEGIN mdsmith merge-driver
 *.md merge=mdsmith
 *.markdown merge=mdsmith
+.git/** -merge
 demo/** -merge
 internal/rules/*/bad/** -merge
 internal/rules/*/bad.md -merge
 internal/rules/*/good/** -merge
 internal/rules/*/fixed/** -merge
+internal/rules/*/pattern/** -merge
 .claude/worktrees/** -merge
+editors/**/node_modules/** -merge
+editors/**/dist/** -merge
 # END mdsmith merge-driver
--- a/.github/workflows/merge-queue.yml
+++ b/.github/workflows/merge-queue.yml
@@ -37,6 +37,14 @@ jobs:
 
       - uses: ./.github/actions/setup-mdsmith-pinned-version
       - name: Install mdsmith merge driver and pre-merge-commit hook
+        # NOTE: switch `merge-driver install` to `merge-driver ci-install`
+        # once the pinned baseline (setup-mdsmith-pinned-version) ships it.
+        # `install` re-renders .gitattributes from .mdsmith.yml; a drifted
+        # committed copy would be rewritten here, dirtying the worktree and
+        # aborting the action's `git merge` — which requeues the PR and
+        # re-fires the labeled trigger forever. `ci-install` verifies
+        # .gitattributes instead of writing it (it cannot dirty the tree)
+        # and fails loudly on drift, turning a loop into one clean red run.
         run: |
           mdsmith merge-driver install
           mdsmith pre-merge-commit install
--- a/cmd/mdsmith/e2e_test.go
+++ b/cmd/mdsmith/e2e_test.go
@@ -1400,6 +1400,65 @@ func TestE2E_MergeDriver_Install(t *testing.T) {
 	assert.NotEmpty(t, hookData, "installed hook must not be empty")
 }
 
+func TestE2E_MergeDriver_CIInstall_InSync(t *testing.T) {
+	dir := t.TempDir()
+	require.NoError(t, exec.Command("git", "init", dir).Run(), "git init")
+
+	// Seed an in-sync .gitattributes plus the driver and hook.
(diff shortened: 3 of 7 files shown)


jeduden:
🟢 **Merge Queue** — picked up

**Next:** No action needed — you'll get another comment when CI starts on the batch. [View merge queue run](https://github.com/jeduden/mdsmith/actions/runs/26837391577).

jeduden:

Merged into batch branch [`merge-queue/batch-456-1780422137`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-456-1780422137). [View CI run](https://github.com/jeduden/mdsmith/actions/runs/26837422941).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`648af0e`](https://github.com/jeduden/mdsmith/commit/648af0e3c9d761399fe6f271be2ade54eed0f6cd). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/26837422941).

PR #460, opened 2026-06-02
test: make per-opt-in-rule time gate runner-independent via a parse ratio

jeduden:
## Problem

The per-opt-in-rule time gate (`TestPerRuleBenchBudget` in `internal/integration/perrule_bench_test.go`) compared an **absolute** `parse+Check` ns/op against fixed per-rule ceilings (1ms; 2ms for MDS035/MDS037). On a contended CI runner the measured time scales with load, so the gate was noise-bound by construction — per-rule ceilings kept having to be bumped (main had just raised MDS043 to 1.5ms) and it still flaked.

## Fix

Gate a **dimensionless ratio** instead of absolute time. `perRuleTiming` times a parse-only op and a parse+Check op **back-to-back each iteration** (a few hundred µs apart — inside one scheduler quantum, so both see the same contention factor `k`), records `(parse+Check)/parse`, and returns the **median over 301 iterations**. `k` cancels *inside* each iteration's ratio, so the number reads the same on a quiet box or a saturated runner. A single uniform `maxTimeRatio = 2.5` replaces the per-rule ns ceilings.

Validated under all-core `yes` load: the no-op ratio holds **0.98–1.01 across a 2.6–4.3× swing in raw parse time** (0/15 runs flake). An earlier design that reduced parse and full *separately* over windows did **not** cancel `k` (the two minima could land in different-`k` windows) — that flaked a no-op at 0.80 on CI; the per-iteration pairing fixes it.

## Measured ratios (`perRuleBenchDoc`)

Every rule but one sits at **0.99–1.28×**; the stable outlier is **MDS043 (no-reference-style) at ~1.65×** — its reference-style link scan is the most Check work relative to parse. That cost is **CPU, not allocation** (MDS043's Check allocates ~10/op). 2.5× clears the outlier with ~50% headroom while binding the cheap rules tighter than the old gate. The deterministic per-rule **allocs** budget stays the tight algorithmic-regression catch; the ratio is the coarse, machine-independent CPU backstop.

## Changes (all in one test file)

- `perRuleTiming` returns `ruleTiming{Ratio, ParseNs, FullNs}` — the **median per-iteration ratio** plus mean times for log context; the gate compares `.Ratio` to `maxTimeRatio`.
- `perRuleBenchBudget` (struct of `{Time, Allocs}`) → `perRuleAllocCeiling` (`map[string]float64`, allocs only). Removed `perRuleCheckNsPerOp`.
- `TestPerRuleTiming` pins the contract with two synthetic rules: no-op Check ⇒ ratio ~1.0; a Check doing one extra `markdown.ParseContext` parse ⇒ ~2.0 (and `> noop`).
- A guard fails fast if a parse op ever measures `0ns`, rather than producing `+Inf`.

No production code changes — entirely the CI gate's measurement.

## Validation

- `TestPerRuleBenchBudget`, `TestPerRuleTiming`, `TestPerRuleBenchDocCompliant` green; full `./internal/integration/` package green
- `golangci-lint` 0 issues; `go vet` clean
- `mdsmith check .` clean

https://claude.ai/code/session_01NY1rwc8LEJ2Ks1GofVPFZA

### DIFF
--- a/internal/integration/perrule_bench_test.go
+++ b/internal/integration/perrule_bench_test.go
@@ -255,7 +255,7 @@ func perRuleTiming(tb testing.TB, r rule.Rule, src []byte, mapFS fstest.MapFS) r
 // perRuleBenchDoc (2026-06-02) every rule but one sits at 0.94-1.33x;
 // the outlier is MDS043 (no-reference-style) at ~1.5-1.65x across runs,
 // whose many allocations charge GC into its Check windows. 2.5x leaves
-// that noisy outlier ~50% headroom yet still binds the 24 cheap rules
+// that noisy outlier ~50% headroom yet still binds the cheap rules
 // tighter than the old absolute gate (1ms vs their ~175µs ≈ 5.7x). The
 // deterministic perRuleAllocCeiling gate below is the tight per-rule
 // algorithmic-regression catch; this is the coarse CPU backstop. If a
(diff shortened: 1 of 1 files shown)


jeduden:
🟢 **Merge Queue** — picked up

**Next:** No action needed — you'll get another comment when CI starts on the batch. [View merge queue run](https://github.com/jeduden/mdsmith/actions/runs/26848764729).

jeduden:

Merged into batch branch [`merge-queue/batch-460-1780435325`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-460-1780435325). [View CI run](https://github.com/jeduden/mdsmith/actions/runs/26848815041).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`b74da69`](https://github.com/jeduden/mdsmith/commit/b74da69a329466d8d78a3c68e400728c1db10338). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/26848815041).

PR #467, opened 2026-06-03
fix(vscode): fix-on-save preview via native editor.codeActionsOnSave (remove mdsmith.fixOnSave)

jeduden:
## The bug

With `mdsmith.previewFix` on, the **Refactor Preview (the diff) never appeared on save**. The server side of plan 207 was correct — the bug was entirely in the extension's fix-on-save wiring.

Fix-on-save used a custom `onWillSaveTextDocument` handler, which drops the preview **three** ways:

1. it pulls edits via `vscode.executeCodeActionProvider`, which **strips the `ChangeAnnotation`/`needsConfirmation` metadata** when the edit crosses the extension-host boundary;
2. it then reduces the result to **bare `TextEdit[]`** via `WorkspaceEdit.entries()`;
3. it applies them inside `event.waitUntil`, which has a **short timeout and cannot host a confirmation UI**.

## What I verified (real headless VS Code)

I stood up a real VS Code via `@vscode/test-electron` and confirmed each link:

- **Server is correct** — a direct LSP probe shows it emits `documentChanges` + `changeAnnotations{needsConfirmation:true}` whenever `previewFix:true` is delivered.
- **VS Code honours `needsConfirmation`** — a hand-crafted annotated edit opens the preview (control test).
- **`executeCodeActionProvider` strips the metadata** — the same edit obtained that way applies immediately, which is exactly why the willSave path can never preview.
- **Native `editor.codeActionsOnSave` previews on save** — applied through the bulk-edit service, which honours the annotation.

## The fix

Remove the custom willSave handler. Fix-on-save is now configured the **ESLint way** — VS Code's native `editor.codeActionsOnSave` with `source.fixAll.mdsmith` (already provided by the LSP):

```jsonc
{ "editor.codeActionsOnSave": { "source.fixAll.mdsmith": "explicit" } }
```

Saves run through the bulk-edit service, so:

- `previewFix` **on** → save opens the **Refactor Preview (the diff)**
- `previewFix` **off** → silent fix

The lightbulb quick-fix path already used native apply and is unchanged.

## `mdsmith.fixOnSave` → deprecated alias (the Prettier way)

`mdsmith.fixOnSave` no longer does anything, but rather than hard-removing it (which would leave existing users with a bare "unknown setting" warning), it's kept as a **deprecated** setting. VS Code renders it struck-through in the Settings UI with a `deprecationMessage`/`markdownDeprecationMessage` pointing to `editor.codeActionsOnSave` — the same way Prettier deprecates settings, with **no runtime popup**.

## Regression guard

Adds a headless VS Code integration harness (`bun run test:e2e`) that drives a real editor and asserts the behaviour end-to-end — the guard the feature lacked (prior tests only checked the server's JSON, never that the editor renders a diff):

```
[CAOS off] outcome=saved   changed=true     # previewFix off -> silent fix
[CAOS on]  outcome=timeout changed=false    # previewFix on  -> preview, buffer untouched
```

## Verification

- `bun test` (extension unit suite): 135 pass
- `bunx tsc --noEmit` (main + e2e tsconfigs): clean
- `bun run test:e2e` (real headless VS Code): pass
- `mdsmith check .`: 405 files, 0 failures

## Docs

`docs/guides/editors/vscode.md`, `docs/guides/install.md`, and the extension `README.md` all document `editor.codeActionsOnSave` as the fix-on-save mechanism; `mdsmith.previewFix` controls whether the diff is shown first.

## Note on the lightbulb

The lightbulb quick-fix applies through VS Code core (the same path this save fix uses), which previews on `needsConfirmation` — so it works once `previewFix` is active. If it shows no diff in a given install, the likely cause is a **bundled mdsmith binary that predates plan 207** (it never emits the annotation); the source on this branch is correct.

https://claude.ai/code/session_012tv9vnDbeuJXMynakqgjy6


jeduden:
🟢 **Merge Queue** — picked up

**Next:** No action needed — you'll get another comment when CI starts on the batch. [View merge queue run](https://github.com/jeduden/mdsmith/actions/runs/26899638562).

jeduden:

Merged into batch branch [`merge-queue/batch-467-1780505477`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-467-1780505477). [View CI run](https://github.com/jeduden/mdsmith/actions/runs/26899676205).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`68c7b1f`](https://github.com/jeduden/mdsmith/commit/68c7b1f8009f15f526e71c659f01a09f712ef22f). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/26899676205).

PR #482, opened 2026-06-05
Windows install: concrete picker command, manual-first guide, Scoop/WinGet plan

jeduden:
## Why

Windows ships a **single** asset, `mdsmith-windows-amd64.exe`, so the
generic GitHub-release one-liner in the install picker —
`curl -LO …/mdsmith-<os>-<arch>` — isn't runnable on Windows. The
guide's Windows steps were also PowerShell-first, when most Windows
users reach for a manual download. And Windows has no package-manager
channel at all.

## What

**1. Concrete Windows command in the install picker — no placeholder.**
- New optional `command-windows` field in the channel SSOT (schema
  `proto.md` + `Channel` struct + generated `channels.yaml`).
- `github-releases.md` sets it to a concrete
  `Invoke-WebRequest …/mdsmith-windows-amd64.exe -OutFile mdsmith.exe`.
- The picker JS swaps to it when the **Windows** chip is active, back
  to the `curl` default otherwise. With JS off, the default still
  renders (no regression).

**2. Manual-first Windows install guide.**
The `### Windows` section now leads with the manual path — browser
download, which folder to keep `mdsmith.exe` in
(`%LOCALAPPDATA%\Programs\mdsmith`), and the Settings UI to add it to
`PATH` — plus a `certutil` verify step that works in Command Prompt or
PowerShell. The PowerShell script stays as the scripted alternative.

**3. Tracked plan for Windows package managers (`plan/234`).**
Scopes Scoop + WinGet distribution, mirroring the Homebrew tap / asdf
plugin pattern (checksum-verified prebuilt `.exe`, a best-effort
release-bump job, generated channel docs that feed the picker/table).
Chocolatey is deferred. No publishing infra created yet — this is the
plan only.

**4. Obsidian guide clarification.**
The Obsidian install has no mac/linux-specific steps to mirror: the
plugin is one `.wasm` file with no native binary or PATH, so the vault
path is identical on every OS. Added a tip to open the plugins folder
from Settings, which sidesteps the hidden `.obsidian` folder (a Windows
papercut).

## Verification

- `go test ./...` green; new TDD for the channel field (round-trip +
  `omitempty`).
- `mdsmith check .` → 414 files, 0 failures; `sync-channels --check`
  clean; `golangci-lint` 0 issues.
- Hugo render confirms the GitHub Releases row carries
  `data-cmd-windows="Invoke-WebRequest …mdsmith-windows-amd64.exe…"`.

[...]

PR #488, opened 2026-06-05
Harden the install-picker channel loader

jeduden:
Follow-up to the install-picker render probe that landed in #482. A post-merge `xhigh` code-review pass surfaced two minor hardening items in `LoadChannelsFromDataFile` (neither a live bug — production `channels.yaml` has 14 channels). Split out here rather than disrupting #482's merge-queue batch.

## Changes

**1. Reuse `channelsDataPath(root)`** instead of a one-off `filepath.Join(root, ChannelsDataFile)`. This matches the sibling readers/writers (`WriteChannelsData`, `CheckChannelsData`) and their `filepath.FromSlash` separator handling, so all four agree on how the slash-separated constant maps to a path.

**2. Error on zero channels.** An empty or truncated `channels.yaml` unmarshals to a nil slice with **no error**. Without a guard, `verify-install-picker` would then expect zero install rows and **pass vacuously on a broken picker** — the exact "broken picker ships green" regression the deploy-job probe exists to catch. Now the loader fails loudly, and the CLI handler surfaces it as a non-zero exit. (`channels-drift` already gates the file, so this is belt-and-suspenders.)

A new test covers the empty-file guard.

## Review

Three review→fix iterations (`xhigh`) converged: iteration 1 found these two items (plus two non-actionable: an unreachable `html.Parse` error branch and a non-constructible `TrimSpace` asymmetry); iterations 2 and 3 found nothing. The empty-set guard was confirmed to be at the right layer — the only caller of `VerifyInstallPicker` reaches it through this loader, so a second guard would be untestable redundancy.

## Verification

- `go test ./...` green · `go tool golangci-lint run` 0 issues · `mdsmith check .` 415/0

https://claude.ai/code/session_01RGBUkRscaPiT5ZfohXWUgL

---
_Generated by [Claude Code](https://claude.ai/code/session_01RGBUkRscaPiT5ZfohXWUgL)_

### DIFF
--- a/internal/release/channels.go
+++ b/internal/release/channels.go
@@ -304,14 +304,19 @@ func CheckChannelsData(root string, chs []Channel) (bool, error) {
 // Hugo template consumed, so a render probe compares the page against
 // its true input rather than a parallel source.
 func LoadChannelsFromDataFile(root string) ([]Channel, error) {
-	path := filepath.Join(root, ChannelsDataFile)
-	data, err := os.ReadFile(path)
+	data, err := os.ReadFile(channelsDataPath(root))
 	if err != nil {
 		return nil, fmt.Errorf("read %s: %w", ChannelsDataFile, err)
 	}
 	var chs []Channel
 	if err := yaml.Unmarshal(data, &chs); err != nil {
 		return nil, fmt.Errorf("parse %s: %w", ChannelsDataFile, err)
+	}
+	// An empty or truncated data file unmarshals to a nil slice with
+	// no error; without this guard a render probe would expect zero
+	// install rows and pass vacuously on a broken picker.
+	if len(chs) == 0 {
+		return nil, fmt.Errorf("%s has no channels", ChannelsDataFile)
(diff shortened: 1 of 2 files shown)


jeduden:
🔍 **Merge Queue** — bisecting

A larger batch failed CI. Bisection is isolating the culprit: this run tests up to **1 of 1** candidate PRs on [`merge-queue/batch-bisect-488-1780678971`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-bisect-488-1780678971). [View current bisect CI run](https://github.com/jeduden/mdsmith/actions/runs/27028686232).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`89b8109`](https://github.com/jeduden/mdsmith/commit/89b8109735e6f6b427b6760c983b4796662ff4d8). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/27028686232).

PR #495, opened 2026-06-05
perf: eliminate per-line string allocs and unsized intermediate slices

jeduden:
## Summary

Audit of the codebase against [`docs/development/high-performance-go.md`](docs/development/high-performance-go.md) identified five categories of allocation anti-patterns in hot paths. This PR fixes all of them, with three additional fixes from code review passes.

### Issues fixed (original top 5)

**1. Fix() methods allocate N strings then join (MDS006, MDS007, MDS008)**

`notrailingspaces`, `nohardtabs`, and `nomultipleblanks` all followed the same pattern:
- Collect `[]string` by calling `string(line)` on every `[]byte` line → N heap allocations
- `strings.Join(..., "\n")` → one more allocation  
- `[]byte(...)` convert back → another allocation

Replaced with `[][]byte` + `bytes.ReplaceAll`/`bytes.TrimRight` + `bytes.Join`. Escape analysis confirms `[]byte("...")` literals are inlined with zero-copy treatment (`go build -gcflags="-m=2"` shows "zero-copy string→[]byte conversion"). Removes the unused `strings` import from all three packages.

**2. `propernames` scan helpers allocate an intermediate slice per AST node**

`scanLines`, `scanCodeSpanChildren`, and `scanRawHTMLSegments` each declared their own `var out []wrongMatch` and returned it, which `collectMatches` then spread-appended into `all`. One extra allocation per block node visited. Refactored to accept the accumulation slice as a parameter and return it — the caller owns the backing array throughout the walk.

**3. `ambiguousemphasis` Check and scanLine**

Pre-sized `diags` in `Check()` to `len(f.Lines)/4+1` to reduce reallocs on emphasis-heavy files. `scanLine` correctly uses `var` nil slices (not `make`) — nil slices don't allocate until first append, while `make([]T, 0, 4)` would heap-allocate on every line even for plain text.

**4. `lsp/patterns` handleRulePatterns — exact-sized slice**

Two-pass: count rules with non-nil `Maintainability` first, then `make([]rulePattern, 0, n)` — avoids the ~10x over-allocation that `len(all)` caused.

**5. `githooks` DiscoverFiles — single `rule.All()` call**

Assign `allRules := rule.All()` once and use for both `len()` and the range loop, avoiding a redundant traversal.

### Code review findings addressed (3 additional fixes)

- **Review pass 1**: Reverted `make([]emphRun, 0, 4)` in `scanLine` — escape analysis confirmed both slices escaped to heap unconditionally, causing 2 allocations per line even with no emphasis characters
- **Review pass 2**: Verified `[]byte("...")` inline literals are zero-copy after inlining; removed redundant package-level `tabBytes`/`spacesBytes` vars from nohardtabs
- **Review pass 3**: Fixed `lsp/patterns` to use an exact-sized slice via count-first pass

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./...` passes (all packages green)
- [x] `go run ./cmd/mdsmith check .` reports 0 failures
- [x] Three xhigh code review passes completed, all findings addressed
- [x] Codecov: all modified lines covered, project coverage at 99.51%

https://claude.ai/code/session_014CdNuuPAmTBgr5cp6rqamV

### DIFF
--- a/internal/githooks/githooks.go
+++ b/internal/githooks/githooks.go
@@ -76,7 +76,7 @@ func ResolveHooksDir(repoRoot string) string {
 // whether to apply a fallback (the install commands do; the
 // git-hook-sync rule does not).
 func DiscoverFiles(repoRoot string, maxBytes int64) []string {
-	directiveNames := make([]string, 0)
+	directiveNames := make([]string, 0, len(rule.All()))
 	for _, r := range rule.All() {
 		if d, ok := r.(gensection.Directive); ok {
 			directiveNames = append(directiveNames, d.Name())
--- a/internal/githooks/githooks.go
+++ b/internal/githooks/githooks.go
@@ -76,8 +76,9 @@ func ResolveHooksDir(repoRoot string) string {
 // whether to apply a fallback (the install commands do; the
 // git-hook-sync rule does not).
 func DiscoverFiles(repoRoot string, maxBytes int64) []string {
-	directiveNames := make([]string, 0, len(rule.All()))
-	for _, r := range rule.All() {
+	allRules := rule.All()
+	directiveNames := make([]string, 0, len(allRules))
+	for _, r := range allRules {
 		if d, ok := r.(gensection.Directive); ok {
 			directiveNames = append(directiveNames, d.Name())
 		}
--- a/internal/lsp/patterns.go
+++ b/internal/lsp/patterns.go
@@ -19,7 +19,7 @@ type rulePattern struct {
 // the same response shape as a workspace where every rule is `maintainability: null`.
 func (s *Server) handleRulePatterns(msg *requestMessage) {
 	all, _ := rules.ListRules()
-	out := make([]rulePattern, 0)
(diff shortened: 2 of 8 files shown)


jeduden:
🔍 **Merge Queue** — bisecting

A larger batch failed CI. Bisection is isolating the culprit: this run tests up to **1 of 1** candidate PRs on [`merge-queue/batch-bisect-495-1780698298`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-bisect-495-1780698298). [View current bisect CI run](https://github.com/jeduden/mdsmith/actions/runs/27043293070).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`0c385b3`](https://github.com/jeduden/mdsmith/commit/0c385b3a600813ff02553faba6137ad5edb1ddae). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/27043293070).

PR #496, opened 2026-06-05
perf: cut GC cost on check/fix — hoist config resolution, pre-size reads, tune batch GOGC

jeduden:
## Why

Instruction-level profiling of a `check` over a neutral corpus (vs. a Rust peer linter at ~the same work) showed mdsmith executing ~2.8× more instructions for the same files — and **~40% of them were Go runtime + GC** (`scanobject`, write barriers, `findObject`), driven by allocation volume, not heap size. The arena fork already cut parse-time allocations; this PR goes after the remaining per-file allocation churn and the GC frequency that scanning it triggers.

## What

Four compounding changes, each measured:

1. **`bytelimit`: pre-size the read buffer.** `ReadFileLimited` already stats the file for its error message — reuse that size to seed the buffer like `os.ReadFile`, instead of `io.ReadAll`'s repeated grow-and-copy. A 1 MB read drops from ~32 allocations to ~3. (`io.ReadAll` was ~22% of allocated bytes.)
2. **`config`: `EffectiveSignature` + `EffectiveAllForKinds`.** `EffectiveAll` depends on `filePath` only through the resolved kind list and the overrides matching it; the signature keys on exactly those, so the engine can memoize.
3. **`engine`: hoist per-run rule work + memoize effective config.** The per-file loop rebuilt the rule→category lookup and re-filtered config-target rules for *every file*, and re-resolved + deep-cloned the effective config per file even when the whole corpus shares one config. Now: lookup hoisted once per run, markdown-rule filter once per worker, and effective configs memoized on a shared, concurrency-safe cache keyed by config signature.
4. **CLI: raise GOGC for batch `check`/`fix`.** A short-lived batch process doesn't need to collect on every heap doubling. `GOGC=400` (still capped by the existing 512 MB `SetMemoryLimit`) trades a little peak memory for less wall time. Respects an explicit `GOGC`; the LSP server and the public library keep the runtime default.

## Results

Parity corpus (234 files, best of 15); allocations over 5616 files:

| | Wall | vs baseline |
|---|---:|---:|
| baseline (main) | 61 ms | — |
| alloc cuts only (1–3, `GOGC=100`) | 47 ms | −23% |
| **all four** | **37 ms** | **−39%** |

- **Allocated bytes: 873.7 → 648.4 MB (−26%)**
- **Allocations: 902,403 → 631,113 (−30%)**

## Correctness

Output is unchanged — the integration **golden tests pass byte-for-byte**, the alloc-budget and performance-gate tests hold, the engine passes `-race` (shared cache is `sync.Map`), and `golangci-lint` is clean. The memo key is provably sufficient: `filePath` enters `EffectiveAll` only via resolved kinds + matching overrides, and the effective map is read-only downstream (`classifyRules` clones a rule's settings before mutating). New code is covered by dedicated unit tests (`bytelimit` alloc bound, `config` signature/equivalence, CLI GC-target decision).

---
_Generated by [Claude Code](https://claude.ai/code/session_01FLNpDmsk8rRuhwtvmvPUJp)_

### DIFF
--- a/cmd/mdsmith/check.go
+++ b/cmd/mdsmith/check.go
@@ -34,6 +34,7 @@ func runCheck(args []string) int {
 	if code >= 0 {
 		return code
 	}
+	tuneGCForBatch()
 	if hasStdin {
 		return checkStdin(opts)
 	}
--- a/cmd/mdsmith/fix.go
+++ b/cmd/mdsmith/fix.go
@@ -25,6 +25,7 @@ func runFix(args []string) int {
 	if code >= 0 {
 		return code
 	}
+	tuneGCForBatch()
 	if hasStdin {
 		fmt.Fprintf(os.Stderr, "mdsmith: cannot fix stdin in place\n")
 		return 2
--- a/cmd/mdsmith/gctune.go
+++ b/cmd/mdsmith/gctune.go
@@ -10,8 +10,9 @@ import (
 // live heap doubles; on a workspace walk that means re-scanning the
 // pointer-rich AST set constantly (profiling attributes ~40% of executed
 // instructions to GC). A batch process is about to exit, so a laxer
-// target trades a little peak memory — already bounded by the 512 MB
-// SetMemoryLimit in run() — for markedly less wall time. 400 was the
+// target trades a little peak memory — already bounded by the soft
+// memory limit in run() (512 MB by default, or the user's GOMEMLIMIT) —
+// for markedly less wall time. 400 was the
 // measured sweet spot on the parity corpus; GOGC=off was slower because
 // an unbounded heap wrecks cache locality.
 const batchGCPercent = 400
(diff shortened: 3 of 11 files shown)


jeduden:
🔍 **Merge Queue** — bisecting

A larger batch failed CI. Bisection is isolating the culprit: this run tests up to **1 of 1** candidate PRs on [`merge-queue/batch-bisect-496-1780698899`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-bisect-496-1780698899). [View current bisect CI run](https://github.com/jeduden/mdsmith/actions/runs/27043663797).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`d4dfcda`](https://github.com/jeduden/mdsmith/commit/d4dfcda1f5bace9e5031d5355c39d79f803a41e4). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/27043663797).

PR #497, opened 2026-06-05
Auto-discover .mdsmith.yml at the vault root in the Obsidian plugin

jeduden:
## Problem

The Obsidian plugin didn't find `.mdsmith.yml`. It only loaded a config when the user explicitly set the **Config path** setting. With it empty (the default), `loadConfigYAML()` returned `""` immediately and the engine ran on built-in defaults — so a `.mdsmith.yml` dropped into the vault was silently ignored.

The setting's own description even claimed to "override the **auto-discovered** `.mdsmith.yml`" — but that discovery was never implemented.

## Fix

Add vault-root auto-discovery. A new `editors/obsidian/src/config.ts` reads `.mdsmith.yml` from the vault root through the `DataAdapter`, which sees dotfiles that Obsidian hides from its vault file API (the same adapter the explicit-path read already uses, so it works on mobile too).

- When **Config path** is empty, `loadConfigYAML()` now falls back to discovery instead of returning `""`.
- An absent or unreadable file still degrades to the engine defaults rather than throwing (an `exists` gate avoids exception-as-control-flow; a post-`exists` read failure is caught).
- An explicit **Config path** continues to take precedence.

This mirrors the CLI, which walks up to the nearest `.mdsmith.yml`. A vault has one root the adapter is rooted at and one lint session over it, so the analog is a single config at the vault root.

## Tests

- **Unit** (`config.test.ts`): `discoverConfigYAML` — present, absent, the exists-gate (no read when absent), read-failure degradation, custom name.
- **Unit** (`main.test.ts`): `loadConfigYAML` resolution — discover when empty, fall back to `""`, explicit-path precedence.
- **End-to-end** (`plugin.e2e.test.ts`): boots the **real WASM engine** with a vault-root `.mdsmith.yml` that disables `line-length`, and asserts the long line no longer reports `MDS001` — proving the config is found **and compiled**, not merely read. (Criterion 3 already proves the same line trips `MDS001` under defaults, so the contrast is airtight.)

`bunx tsc --noEmit` clean, all 82 plugin tests pass, and `mdsmith check .` passes (418 files).

## Docs / UX

Updated the Obsidian guide (install + settings table + troubleshooting) and the settings UI copy/comments to describe auto-discovery.

https://claude.ai/code/session_012oegF5TA5sxAJgfBoPnmwL

---
_Generated by [Claude Code](https://claude.ai/code/session_012oegF5TA5sxAJgfBoPnmwL)_

### DIFF
--- a/docs/guides/editors/obsidian.md
+++ b/docs/guides/editors/obsidian.md
@@ -137,7 +137,11 @@ in `<vault>/.obsidian/plugins/mdsmith/` — `main.js`,
 `manifest.json`, `styles.css`, `mdsmith.wasm`, and
 `wasm_exec.js`. A missing `mdsmith.wasm` or
 `wasm_exec.js` is the usual cause. Re-unzip the
-release and reload the plugin.
+release and reload the plugin. If the notice instead
+reads `parsing config file: …`, your `.mdsmith.yml`
+has a syntax error — the engine refuses an invalid
+config rather than linting on defaults — so fix the
+YAML and run `mdsmith: Restart session`.
 
 **No diagnostics appear.** Check that **Run mode** is
 not `off`. With `onSave`, diagnostics refresh only
--- a/docs/guides/editors/obsidian.md
+++ b/docs/guides/editors/obsidian.md
@@ -100,8 +100,10 @@ drop the unzipped files there.
 You need Obsidian 1.5 or later. A config file is
 optional: mdsmith lints with built-in defaults, so the
 plugin works as soon as you enable it. To tune the
-rules, add a `.mdsmith.yml` to the vault and point the
-**Config path** setting at it.
+rules, add a `.mdsmith.yml` to the vault root and the
+plugin discovers it automatically — no setting to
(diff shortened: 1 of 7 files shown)


jeduden:
🔍 **Merge Queue** — bisecting

A larger batch failed CI. Bisection is isolating the culprit: this run tests up to **1 of 1** candidate PRs on [`merge-queue/batch-bisect-497-1780701042`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-bisect-497-1780701042). [View current bisect CI run](https://github.com/jeduden/mdsmith/actions/runs/27044900225).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`d6b0b7d`](https://github.com/jeduden/mdsmith/commit/d6b0b7da09264680d15b72fdd3ab78e6ee06c8df). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/27044900225).

PR #500, opened 2026-06-05
ci(benchmark): publish rendered benchmark pages to the assets branch

jeduden:
## What

On every merge to `main`, render the benchmark Markdown pages — the top-level `README.md` (headline) and the benchmark `README.md` (result tables) — from main's sources with the freshly measured numbers spliced in, and publish them to the orphan **`assets` branch** alongside the raw numbers (`data/` + fragments).

## Why this shape

The original idea was to commit refreshed numbers back to `main`, but a 3-agent review found a hard blocker: **the default `GITHUB_TOKEN` cannot push to protected `main`** (proven by `merge-queue.yml` needing a dedicated bypass PAT). Working around it would have required a bypass PAT → which re-triggers the workflow (loop) → `[skip ci]` → and a build/push split to contain the bypass-token blast radius.

Publishing to **`assets`** sidesteps all of it: the `publish` job already has write access there, so there's **no protected-main push, no bypass token, no trigger loop, and no `set -e`/branch-protection hazards.**

## How

- **`record` job (creds-free):** after the bench, `mdsmith fix README.md docs/research/benchmarks/README.md` splices the fresh fragments into the page bodies, and the rendered pages are staged into the artifact under `pages/`. The render runs here precisely because this job holds **no push token** — addressing the review's "don't compile/run under a push token" concern.
- **`publish` job (git-only):** unchanged — it already copies the artifact subtree to `assets/benchmarks/`, so `pages/README.md` and `pages/benchmark.md` ride along.
- The in-repo committed snapshot on `main` is **untouched** — it stays the `bench-fragments` gate's source of truth, hand-refreshed via `run.sh`.

## Review findings, resolved

This supersedes the commit-to-main implementation (force-pushed over). The new shape moots the review's blocker (protected-main push), the `set -e`-in-`until` correctness bug, the bypass-token/compile-under-token security delta, and the phantom-`markdown-linters.md` issue (the render fixes only the two real includers).

## Validation

YAML parses (`record` + `publish` only); `go build` clean; the render step is a no-op exit-0 on in-sync fragments; `mdsmith check .` passes (422 files).

> Note: the branch name `benchmark-commit-to-main` is now a misnomer — the change publishes to `assets`, not main.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -76,7 +76,7 @@ jobs:
              docs/research/benchmarks/headline.fragment.md \
              artifacts/benchmarks/
           cp docs/research/benchmarks/data/*.json artifacts/benchmarks/data/
-          cp README.md artifacts/benchmarks/pages/README.md
+          cp README.md artifacts/benchmarks/pages/repo-readme.md
           cp docs/research/benchmarks/README.md \
              artifacts/benchmarks/pages/benchmark.md
       - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,12 +1,14 @@
 name: Cross-tool benchmark
 
 # Re-measure the mdsmith-vs-Rust benchmark on every merge to
-# main and publish the refreshed numbers to the orphan `assets`
-# branch (the demo.gif pattern). Record-only: it never gates a
-# merge or a release. The website pulls these numbers at
-# build time; the in-repo committed snapshot stays the
-# `bench-fragments` CI gate's source of truth and is unaffected
-# by this workflow.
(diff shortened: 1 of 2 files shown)


jeduden:
🔍 **Merge Queue** — bisecting

A larger batch failed CI. Bisection is isolating the culprit: this run tests up to **1 of 1** candidate PRs on [`merge-queue/batch-bisect-500-1780707060`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-bisect-500-1780707060). [View current bisect CI run](https://github.com/jeduden/mdsmith/actions/runs/27047828808).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`b93868b`](https://github.com/jeduden/mdsmith/commit/b93868be645701b6c63e4faa50c25a0d03ef760f). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/27047828808).

PR #505, opened 2026-06-06
refactor(gctune): single-source the batch GC default; apply it in WASM too

jeduden:
## What

Make the batch GC default (`GOGC`-equivalent 400, from #496) a **single source of truth** and apply it in the **WASM build** too.

- New **`internal/gctune`** package: `BatchPercent = 400`, `Target(env)` (pure, testable), `ApplyBatch()`.
- `cmd/mdsmith` `check`/`fix` now call `gctune.ApplyBatch()` (the local `cmd/mdsmith/gctune.go` is removed).
- `cmd/mdsmith-wasm` `main()` now calls `gctune.ApplyBatch()` — previously the WASM engine never applied the bump at all.
- The LSP server and the embedded `pkg/mdsmith` library still **don't** call it, so long-running/embedded callers keep Go's default.

## Still env-free

`ApplyBatch()` sets `debug.SetGCPercent(400)` programmatically — no `GOGC` env var required. An explicitly-set `GOGC` is honored as the opt-out (`Target` returns -1).

## ⚠️ Worth a look: WASM is long-lived

The batch bump was designed for **short-lived** processes (CLI exits, OS reclaims). The WASM engine is a **long-lived session** (the Obsidian plugin keeps it loaded) — the same category as the LSP we deliberately *excluded*. A high `GOGC` permanently relaxes GC for the session, and WASM linear memory grows but doesn't shrink. In practice the per-lint live set is small (a few MB peak) and the latency win on interactive linting is real, so I applied it at startup per the "single source of truth" goal. If the idle-session memory cost matters (esp. mobile), the alternatives are: scope it to *during* each check/fix call, or exclude WASM like the LSP. Easy follow-up either way.

## Validation

Native build, **WASM compile** (`GOOS=js GOARCH=wasm`), `go vet` (incl. WASM), `golangci-lint` (0 issues), and the **full test suite** all pass. Behavior is unchanged for the CLI; the only new behavior is the WASM build applying the default.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FLNpDmsk8rRuhwtvmvPUJp)_


jeduden:
🔍 **Merge Queue** — bisecting

A larger batch failed CI. Bisection is isolating the culprit: this run tests up to **1 of 1** candidate PRs on [`merge-queue/batch-bisect-505-1780735998`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-bisect-505-1780735998). [View current bisect CI run](https://github.com/jeduden/mdsmith/actions/runs/27057947347).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`5c354b8`](https://github.com/jeduden/mdsmith/commit/5c354b80a26a54f8f880005edbf20724d5866911). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/27057947347).

PR #510, opened 2026-06-06
Fix WinGet manifest (portable) and delist it until live

jeduden:
## Why

Two problems with the WinGet channel:

1. **The manifest was wrong.** `render-winget-manifest` declared `InstallerType: exe` with `Silent: /S`, but `mdsmith-windows-amd64.exe` is a bare CLI binary, not an installer. WinGet would run it as an installer (`mdsmith-windows-amd64.exe /S`) — the linter rejects `/S` and exits non-zero, so nothing lands on PATH. Microsoft's sandbox validation would likely reject the PR too. `komac update` inherits the bootstrap manifest's installer structure, so this would propagate to every release.
2. **It was advertised before it works.** The install picker and table listed `winget install jeduden.mdsmith`, which does nothing until the `microsoft/winget-pkgs` manifest PR merges and clears moderation. (Scoop stays listed — its bucket repo is available; asdf/mise stay because their listed command works today via an alternate form.)

## What

**Commit 1 — portable manifest.** `wingetInstallerManifest` now emits `InstallerType: portable` with `PortableCommandAlias: mdsmith`, dropping the installer-only `InstallerLocale`/`InstallModes`/`InstallerSwitches`. WinGet stores the binary and links it onto PATH as `mdsmith`.

**Commit 2 — delist WinGet.** New optional channel frontmatter flag `unlisted` (default `false`):
- `sync-channels` skips unlisted channels when projecting `website/data/channels.yaml` (the picker).
- The install-guide table catalog excludes `winget.md` by glob (the catalog's CUE `where` filter would otherwise require *every* channel to declare the field, since `Match` requires referenced paths to exist).
- `winget.md` sets `unlisted: true`; the WinGet block is removed from the Windows section of the install guide.

The channel doc and release tooling stay — only the user-facing listings wait for the manifest PR. `plan/234` is updated to match.

## Tests

- Flipped `TestRenderWingetManifest_InstallerFields` to assert `portable` + `PortableCommandAlias` and the absence of the exe/silent fields (red → green).
- Added `TestLoadChannels_SkipsUnlisted` (red → green).
- `go test ./...`, `go vet ./...`, `go tool golangci-lint run` (0 issues), `mdsmith check .`, and `sync-channels --check` all pass.

## Still external (unchanged, pre-existing)

`winget install jeduden.mdsmith` only works once the bootstrap manifest PR is merged into `microsoft/winget-pkgs` and clears moderation. When that lands, drop `unlisted: true` and the table glob exclusion to re-list it.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FuwEfh9fYYzXrbX1vZ6i8F)_


jeduden:
🔍 **Merge Queue** — bisecting

A larger batch failed CI. Bisection is isolating the culprit: this run tests up to **1 of 1** candidate PRs on [`merge-queue/batch-bisect-510-1780752678`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-bisect-510-1780752678). [View current bisect CI run](https://github.com/jeduden/mdsmith/actions/runs/27063649228).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`bb85e79`](https://github.com/jeduden/mdsmith/commit/bb85e79716039628835ce2b1fd4d99ea075c598e). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/27063649228).

PR #534, opened 2026-06-09
perf: eliminate per-call allocations in 5 hot-path functions

jeduden:
## Summary

Audited the codebase against `docs/development/high-performance-go.md` and identified five functions that allocate on every invocation due to `strings.Split`, `bytes.Split`, or `strings.Join` calls that can be replaced with zero-allocation alternatives. Each fix follows Red/Green TDD: a failing alloc test was added first, then the implementation was changed to make it pass.

### Fixes

| # | Function | File | Anti-pattern | Allocs before → after |
|---|----------|------|--------------|----------------------|
| 1 | `containsDotDotElement` | `internal/rules/include/rule.go` | `strings.Split(p, "/")` in hot per-directive path | 1 → **0** |
| 2 | `minFenceLen` | `internal/rules/include/rule.go` | `strings.Split(text, "\n")` allocates a full lines slice | 1 → **0** |
| 3 | `paragraphEndLine` | `internal/rules/noreferencestyle/rule.go` | `bytes.Split(source, "\n")` re-split the whole file; caller now passes `f.Lines` directly | 1 → **0** |
| 4 | `matchDoublestar` | `internal/gitignore/gitignore.go` | `strings.Split(pattern, "**")` + `strings.Split(path, "/")` + `strings.Join` inside a loop — O(n) allocations per path | 3–5 → **0** |
| 5 | `collectBodySyncPoints` | `internal/rules/requiredstructure/rule.go` | `bytes.Split(content, "\n")` on schema body | 3 → **2** (only the inherent `string()` casts for heading comparison remain) |

### Motivation

`docs/development/high-performance-go.md` specifies: _"each removed alloc in a per-file Check saves one alloc per workspace file"_ and calls out `strings.Split`/`bytes.Split` returning newly-allocated slices as the canonical anti-pattern to avoid. With a 600-file corpus, each 1-alloc reduction translates directly to 600 fewer heap allocations per `mdsmith check` run.

`matchDoublestar` is called once per `.gitignore` pattern per file during workspace discovery — the most upstream hot path. Eliminating 3–5 allocations per `**`-pattern match compounds across all workspace files and all rules.

### Approach

- `containsDotDotElement`: replaced with `HasPrefix`/`HasSuffix`/`Contains` checks that match the exact four cases without any slice allocation.
- `minFenceLen`: a single-pass byte scan replaces the split+inner-loop; `\n` is just another non-`` ` `` character that resets the run counter.
- `paragraphEndLine`: changed parameter type from `[]byte` (raw source) to `[][]byte` (pre-split lines from `lint.File`). Updated the single call site to pass `f.Lines`.
- `matchDoublestar`: replaced `strings.Split(pattern, "**")` with `strings.Index` (no alloc); replaced the `strings.Split(path, "/")` + per-iteration `strings.Join` with index-based substringing (Go string slices share the backing array). Extracted three helper functions (`matchSingleDoublestar`, `matchLeadingDoublestar`, `matchMiddleDoublestar`) to keep the cognitive complexity below the 30-node lint ceiling while making each branch readable.
- `collectBodySyncPoints`: replaced `bytes.Split(content, "\n")` with a direct `i`-pointer scan over `content`, eliminating the `[][]byte` header slice. The two `string()` casts for heading-line comparison against `headingMatchesLine` are inherent and unchanged.

## Test plan

- [x] New alloc tests added for all 5 functions (`*_test.go` beside each changed file)
- [x] Alloc tests confirmed **failing before** the fix (red phase)
- [x] Alloc tests confirmed **passing after** the fix (green phase)
- [x] All existing tests in changed packages pass (`go test ./...` — 0 failures)
- [x] `go vet ./...` — clean
- [x] `golangci-lint run` on changed packages — 0 issues
- [x] `mdsmith check .` on repository — 437 files, 0 failures

https://claude.ai/code/session_017d9EhXPVmeoCfMq1747hCU

---
_Generated by [Claude Code](https://claude.ai/code/session_017d9EhXPVmeoCfMq1747hCU)_

### DIFF
--- a/internal/gitignore/alloc_test.go
+++ b/internal/gitignore/alloc_test.go
@@ -1,0 +1,42 @@
+package gitignore
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// TestMatchDoublestar_LeadingDoublestar_ZeroAllocs confirms the
+// leading-** path allocates nothing after the strings.Split/Join rewrite.
+func TestMatchDoublestar_LeadingDoublestar_ZeroAllocs(t *testing.T) {
+	cases := []struct{ pattern, path string }{
+		{"**/*.md", "readme.md"},
+		{"**/*.md", "sub/readme.md"},
+		{"**/*.md", "a/b/c.md"},
+		{"**/foo", "a/b/c/foo"},
+	}
+	for _, c := range cases {
+		allocs := testing.AllocsPerRun(100, func() {
+			matchDoublestar(c.pattern, c.path)
+		})
+		assert.Equal(t, 0.0, allocs,
+			"matchDoublestar(%q, %q) allocs: want 0, got %v", c.pattern, c.path, allocs)
+	}
+}
+
+// TestMatchDoublestar_MiddleDoublestar_ZeroAllocs confirms the
+// middle-** path allocates nothing after the rewrite.
+func TestMatchDoublestar_MiddleDoublestar_ZeroAllocs(t *testing.T) {
+	cases := []struct{ pattern, path string }{
+		{"a/**/b.md", "a/b.md"},
+		{"docs/**/readme.md", "docs/readme.md"},
+		{"docs/**/readme.md", "docs/sub/readme.md"},
+	}
+	for _, c := range cases {
+		allocs := testing.AllocsPerRun(100, func() {
+			matchDoublestar(c.pattern, c.path)
+		})
+		assert.Equal(t, 0.0, allocs,
+			"matchDoublestar(%q, %q) allocs: want 0, got %v", c.pattern, c.path, allocs)
+	}
+}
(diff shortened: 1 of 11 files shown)


jeduden:
🔍 **Merge Queue** — bisecting

A larger batch failed CI. Bisection is isolating the culprit: this run tests up to **1 of 1** candidate PRs on [`merge-queue/batch-bisect-534-1781038706`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-bisect-534-1781038706). [View current bisect CI run](https://github.com/jeduden/mdsmith/actions/runs/27235435621).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`a201e53`](https://github.com/jeduden/mdsmith/commit/a201e530c74d380b713315b075f981df04187483). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/27235435621).

PR #541, opened 2026-06-10
markdownlint migration: auto-conversion via `mdsmith init --from-markdownlint` + guide rewrite

jeduden:
## What

Two halves of one goal — make the markdownlint migration mechanical and keep its docs drift-free:

1. **`mdsmith init --from-markdownlint[=path]`** converts a markdownlint config (`.markdownlint.jsonc/.json/.yaml/.yml`, `.markdownlintrc`; auto-discovered when no path is given) into a minimal `.mdsmith.yml`. Everything it cannot translate faithfully becomes a note — on stderr and as a `# Not converted:` comment block in the generated file. It never guesses settings.
2. **Docs rewrite.** The migration guide becomes a four-step how-to with real command output. Its hand-written rule table — which had drifted (MD046 pointed at `fenced-code-style` instead of `code-block-style`; it documented a `no-duplicate-headings.siblings-only` setting that does not exist) — is replaced by `docs/reference/markdownlint-mapping.md`, a `<?catalog?>` table generated from rule README front matter.

## How drift is prevented

- The MD→MDS rule mapping is **not hand-written**: the converter reads the `markdownlint:` front matter embedded in the rule READMEs (`internal/rules`), the same source the coverage matrix and the new reference table regenerate from.
- The only hand data is the per-rule **option translation table** (e.g. `MD013.line_length` → `line-length.max`), and it is pinned by tests: every translated settings map is applied through the live rules' `ApplySettings`, and the copied `line-length` exclude default is compared against the rule's `DefaultSettings`.

## Conversion semantics

- Explicit keys (MD ids or markdownlint alias names) translate; only behavior-changing entries are emitted, so unmentioned rules keep mdsmith defaults.
- `default: false` sweeps every markdownlint-mapped, default-on mdsmith rule to `false`; mdsmith-only rules are untouched.
- A rule covering several MD ids (e.g. `atx-heading-whitespace` ← MD018–MD021, MD023) stays enabled when only some analogs are disabled, with a note.
- markdownlint-default-on checks whose mdsmith analog is opt-in (`no-inline-html`, `single-h1`, …) are reported once as a review note instead of being silently lost or opinionatedly enabled.

## Tests

- Unit: parser (JSONC comments/trailing commas/BOM/aliases-rejection), converter semantics, option table, emitter (golden output + config round-trip).
- E2E: auto-discovery, explicit path, no-config error, and `mdsmith check` running clean under a converted config.
- `go test ./...` green; `mdsmith check .` green (433 files).

https://claude.ai/code/session_01Um2P9Hxb6CsjuBsuREZeML

---
_Generated by [Claude Code](https://claude.ai/code/session_01Um2P9Hxb6CsjuBsuREZeML)_


jeduden:
🟢 **Merge Queue** — picked up

**Next:** No action needed — you'll get another comment when CI starts on the batch. [View merge queue run](https://github.com/jeduden/mdsmith/actions/runs/27255947306).

jeduden:

Merged into batch branch [`merge-queue/batch-541-1781070279`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-541-1781070279). [View CI run](https://github.com/jeduden/mdsmith/actions/runs/27255970741).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`2bff8ae`](https://github.com/jeduden/mdsmith/commit/2bff8ae06ec2ea29f92338c63acb2fea6154117f). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/27255970741).

PR #544, opened 2026-06-10
Plan 2606100533 + 2606100534: timestamp plan ids and the unique-frontmatter rule

jeduden:
Implements both plans from the original design (the design docs themselves are the first two commits and dogfood the new scheme as their own ids).

**Plan 2606100533 — coordination-free plan ids (✅)**
- Plan ids are now the minute-precision UTC creation timestamp (`date -u +%y%m%d%H%M`); the allocation contract lives in plan/proto.md, and the proto `id:` type freezes the legacy range — `(int & >=1 & <=243) | (int & >=2601010000)` — so a stray `max+1` id fails MDS020.
- The nine duplicate twin pairs are renumbered (keepers chosen by reference weight; all nine moved twins are completed plans; zero `depends-on` edits were needed). All references moved with them, including the shipped `see plan 156` diagnostic strings, three `internal/` Go comments pinning plan 215, and the architecture-audit links. plan/157's orphaned `depends-on: [19]` is deleted.
- The catalog's `numeric:id` sort parses with `ParseInt(..., 64)` so 32-bit targets (the tinygo goal) keep numeric order.
- pick-plan and the architecture audit-checklist drop their max+1 / duplicate-id workarounds.

**Plan 2606100534 — MDS069 unique-frontmatter (🔳, enablement gated)**
- New rule: within configured `include`/`exclude` globs, no two files may share a value in the configured front-matter `field`. Later file in path order gets the diagnostic at its front-matter line, naming the field, value, and first holder. Values compare as canonical scalars (`7`/`"7"` and `0x10`/`16` collide). Inert without config, so it ships enabled by default.
- Value→first-holder index builds once per run in a new `RunCache` slot, keyed by absolute paths so absolute-path args, Windows separators, and subdirectory invocations all resolve; symlinks are skipped per the plan-84 posture. Targeted invalidation (a `ScopeInvalidator` registered post-once, following the `ParsedSchemaMetadata` pattern) keeps the index warm across out-of-scope LSP edits and fails toward a rebuild when a path can't be related to the root.
- **Rollout sequencing**: the `.mdsmith.yml` enablement block is deliberately NOT in this PR. The pinned `mdsmith-fixed-version` binary (v0.33.0) silently ignores the unknown rule key — verified empirically — so early config would enforce uniqueness under the branch binary but not the pinned one. Per docs/development/adopt-new-directive-syntax.md: merge this rule → cut a release → bump `MDSMITH_VERSION` in `setup-mdsmith-pinned-version` → then land the config block. Plan 2606100534 stays 🔳 with that gated task as the tracked reminder.

**Review**: three full xhigh review rounds (8–9 finder angles each) ran against this branch; every confirmed finding is fixed and pinned by tests — among them a path-space mismatch that made the rule silently inert outside `check .`-from-root (absolute, Windows, and subdirectory shapes, each reproduced), symlink read-through, an 8 ms index rebuild on every LSP keystroke, per-Check allocations invisible to both CI gates, and a macOS symlinked-root case that would have frozen stale verdicts for a whole editor session. Race-detector stress runs, adversarial front matter (anchor bombs, size limits), and glob-widening probes are clean; warm `Check` measures 2 allocs/op. Known deliberate gaps are documented in the plan: the index reads as-saved disk state in the LSP (same posture as every cross-file rule; overlay-into-RootFS is engine-level follow-up), and `.mdsmith.yml` rule-name typos are not validated against the registry anywhere (pre-existing, repo-wide).

Side effect: regenerating PLAN.md healed catalog rows that had drifted stale on main (the table was last rendered by an older formatter).

`mdsmith check .` passes on all 438 files; `go test ./...` and `golangci-lint` are clean.

https://claude.ai/code/session_01JASHJfDX5XQy8j4V4ubByz


jeduden:
🔍 **Merge Queue** — bisecting

A larger batch failed CI. Bisection is isolating the culprit: this run tests up to **1 of 1** candidate PRs on [`merge-queue/batch-bisect-544-1781085322`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-bisect-544-1781085322). [View current bisect CI run](https://github.com/jeduden/mdsmith/actions/runs/27268300027).

jeduden:
❌ **Merge Queue** — CI failed (identified via bisection)

Bisection identified this PR as the failing change. [View CI run that isolated the failure](https://github.com/jeduden/mdsmith/actions/runs/27268300027).

**Next:** Fix the failure, push updates, then re-add the `queue` label to retry.

PR #545, opened 2026-06-10
Plan 242: Recover from panics in the LSP lint pipeline

jeduden:
Draft PR for [plan/242_lsp-panic-recovery.md](plan/242_lsp-panic-recovery.md).

Status will move 🔲 → 🔳 in PLAN.md once the first real
commit lands. Marking ready for review when the
acceptance criteria are checked off.

---
_Generated by [Claude Code](https://claude.ai/code/session_019LXhkAzG4UNYyne1Mgczwf)_

PR #552, opened 2026-06-10
lsp: answer panicked requests, surface recoveries, contain the settings goroutine

jeduden:
Follow-up hardening of the panic containment shipped in #545 ([plan/242_lsp-panic-recovery.md](plan/242_lsp-panic-recovery.md), audit finding S001), from its post-merge review.

## What

1. **A panicked request is now answered.** `dispatchRaw`'s recover writes an InternalError (`-32603`) response for the in-flight request ID. Previously the frame was dropped silently: the server survived, but the client's promise (code action, hover, rename…) pended forever — `vscode-languageclient` has no request timeout, so the crash had become a hang. The dispatch panic seam moves after frame parsing, where real handler panics occur, so the test can assert the reply.
2. **Recovery is visible in production.** A shared `logPanic` helper logs the panic with `debug.Stack()` and mirrors the same text to `window/logMessage`. `mdsmith lsp` constructs the server without a `Logger` (disabled by default), so the previous log-only recovery left zero trace in shipped binaries — diagnostics just froze silently.
3. **The third goroutine origin is contained.** `go s.fetchClientSettings(ctx)` runs config load, kind-schema CUE compile, session rebuild, and the host's `OnConfigReload` callback with no recover — the same reload pipeline that was already contained when triggered from the dispatch loop. A deferred `recoverPanic` closes the asymmetry; a hostile workspace config can no longer kill the server via the settings round-trip while surviving via `initialized`.
4. **Comment accuracy.** The lint seam comments claimed the hook "replaces" `CheckVersion` (it fires just before it); a test comment named a nonexistent test and described the InternalError reply as unimplemented — it is now implemented and asserted.

The parent-process and singleton watcher goroutines remain unprotected by choice: they process no untrusted content.

## Tests

All red/green via the existing seams plus the host `OnConfigReload` callback as a real in-frame panic source: the answered-request assertions (`"id":99` + `-32603`), stack-trace presence (`goroutine`), `window/logMessage` surfacing, and `fetchClientSettings` containment. `go test -race ./internal/lsp/`, full `go test ./...`, `golangci-lint`, and `mdsmith check .` all green.

https://claude.ai/code/session_019LXhkAzG4UNYyne1Mgczwf

---
_Generated by [Claude Code](https://claude.ai/code/session_019LXhkAzG4UNYyne1Mgczwf)_

### DIFF
--- a/internal/lsp/panic_recovery_test.go
+++ b/internal/lsp/panic_recovery_test.go
@@ -3,6 +3,8 @@ package lsp
 import (
 	"context"
 	"encoding/json"
+	"os"
+	"path/filepath"
 	"strings"
 	"testing"
 	"time"
@@ -10,6 +12,7 @@ import (
 	vlog "github.com/jeduden/mdsmith/internal/log"
 	"github.com/jeduden/mdsmith/internal/rule"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 // TestRunLintPanicIsRecovered verifies that a panic inside the lint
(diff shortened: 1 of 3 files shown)


jeduden:
🟢 **Merge Queue** — picked up

**Next:** No action needed — you'll get another comment when CI starts on the batch. [View merge queue run](https://github.com/jeduden/mdsmith/actions/runs/27267161202).

jeduden:

Merged into batch branch [`merge-queue/batch-552-1781084101`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-552-1781084101). [View CI run](https://github.com/jeduden/mdsmith/actions/runs/27267196338).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`2ac5ab0`](https://github.com/jeduden/mdsmith/commit/2ac5ab0c1fd3d1af3e1a5ee9f87c2effd5299922). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/27267196338).

PR #576, opened 2026-06-11
feat(tinygo): add build-tagged seams for tinygo-incompatible OS calls (plan 247)

jeduden:
## Summary

- Add `//go:build tinygo` / `//go:build !tinygo` seam pairs for each stdlib call tinygo's wasm target does not implement: `os.Chmod`, `os.SameFile`, `filepath.EvalSymlinks`
- Seven packages gain seam file pairs: `internal/fix`, `internal/schema`, `internal/githooks`, `internal/lsp`, `internal/rules/build`, `internal/rules/crossfilereferenceintegrity`, `internal/rules/requiredstructure`
- Tinygo variants: no-op for chmod (no POSIX mode bits in wasm), identity for EvalSymlinks (no symlinks in the wasm sandbox), `true` for SameFile in githooks (wasm sandbox has no symlinks/concurrent writes so TOCTOU check should always pass), `false` for SameFile in requiredstructure (triggers path-equality fallback in `isSchemaFile`)
- Non-tinygo paths delegate to the real stdlib calls unchanged
- `chmodFile` vars in `internal/fix` and `internal/schema` are injectable via package-level vars; both packages guard them with `chmodFileMu sync.Mutex` so parallel tests don't race
- `isSchemaFile` restructured: `sameFile` result is additive (true → return immediately) rather than replacing the path-equality fallback, which now always runs as a backstop; `filepath.Clean` wrappers on `filepath.Abs` results removed (Abs already cleans)
- `githooks/compat_tinygo.go` `sameFile` returns `true` (not `false`) so `atomicWriteGitattributes` doesn't always fail with "file changed since lstat" when updating an existing file on tinygo
- Un-skip `TestTinyGoWASMArtifactSizeBudget` — now asserts the 8 MiB ceiling rather than skipping
- Flip the `tinygo-wasm` CI job to enforcing by removing `continue-on-error: true`
- Mark plan 247 `✅`; update plan 218, plan 240, and `docs/background/concepts/engine-api.md` to record tinygo as CI-verified
- File three follow-on plans for out-of-scope findings: consolidate 16 compat file pairs into `internal/oscompat` (plan 2606111048), harden WASM size test with `-no-debug` and gzip check (plan 2606111049), mutex gap in schema package (plan 2606111050, already fixed in this PR)

## Test plan

- [x] `go build ./...` passes (the `!tinygo` path compiles cleanly)
- [x] `go test ./...` passes — all tests green, no regressions
- [x] `schema.chmodFile` guarded with `chmodFileMu` matching the `fix` package pattern
- [ ] CI `tinygo-wasm` job runs enforcing (no `continue-on-error`) and `TestTinyGoWASMArtifactSizeBudget` asserts the 8 MiB budget
- [ ] `TestWASMArtifactSizeBudget` (standard GOOS=js build) unchanged
- [ ] Codecov patch coverage gate passes

https://claude.ai/code/session_01B1TZmZyubdqjWrXoNum3dN

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -519,12 +519,8 @@ jobs:
   # same `go test` run also enforces the artifact size budget
   # (size_test.go, a regression guard now at ~11.2 MB raw / ~2.8 MB
   # gzipped after cuelang.org/go left the graph) and the Go↔JS method-set
-  # parity (methods_test.go). tinygo is intentionally not built here: the
-  # tinygo wasm target still leaves os.Chmod, os.SameFile, and
-  # os.Symlink/filepath.EvalSymlinks undefined, which pkg/mdsmith reaches
-  # transitively, so the tinygo build does not yet compile. Build-tagging
-  # those calls out of the wasm graph is follow-up work (see
-  # docs/background/concepts/engine-api.md and plan/240).
+  # parity (methods_test.go). tinygo is built in the separate tinygo-wasm
+  # job below, which is now enforcing (plan 247).
   wasm:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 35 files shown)


jeduden:
🟢 **Merge Queue** — picked up

**Next:** No action needed — you'll get another comment when CI starts on the batch. [View merge queue run](https://github.com/jeduden/mdsmith/actions/runs/27343059115).

jeduden:

Merged into batch branch [`merge-queue/batch-576-1781176760`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-576-1781176760). [View CI run](https://github.com/jeduden/mdsmith/actions/runs/27343137669).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`4a25f37`](https://github.com/jeduden/mdsmith/commit/4a25f37ae35100cdbcdd4c0c99f568294c0f5169). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/27343137669).

PR #592, opened 2026-06-12
Code-review fixes for #591: enforce the soft-skip contract, fix the vs-mado denominator, refresh stale prose

jeduden:
Follow-up to the merged #591: the user-requested xhigh code-review round over that diff surfaced three fixable findings (plus structural notes left for the maintainer). Round 2 of the review runs on this PR next.

## Fixes

**1. Required smoke channels can no longer soft-skip verification** (`internal/release/releasesmoke.go`, TDD)

#591 introduced the `skipped=true` output contract so the best-effort `mise-registry` leg skips the shared Verify step. The review found `CheckReleaseSmoke` only checks that each required channel has a matrix entry — a required channel carrying the soft-skip marker would pass the gate while never verifying the published binary (the concrete risk: promoting `mise-registry` into `RequiredSmokeChannels` once the jdx registry entry lands, without removing its skip path). The gate now rejects the combination; tests cover both the violation and the allowed best-effort case.

**2. The "vs mado" benchmark column now actually divides by mado** (`docs/research/benchmarks/gen_fragments.py`)

The table's ratio denominator was `min()` over all tools under a hardcoded `vs mado` header. In the pre-#591 committed data mdsmith-parity was faster than mado, so every published ratio was silently vs-parity — the old table literally showed mado at "1.1x vs mado". Pinned to mado's median with a loud exit if mado's result is missing; output is byte-identical for the current data except the caption, which now reads "the ratio to mado's median". (The headline's `int()` floor was left alone — its comment states the floor is deliberate: "the headline must never overstate the win".)

**3. Stale hand-written README numbers** (`docs/research/benchmarks/README.md`)

The corpora table and the growth paragraph still said 722 repo files (the fragments say 766); the parity-gap sentence said "trails mado by roughly a tenth" while the data gives 1.14x — it now cites the table's own 1.1x figure so it can't drift independently. Also added one clause to the Verify-step comment in `release.yml`: the skip mechanism relies on the implicit `success()` guard, so `continue-on-error` must not be added to the Install step.

## Review findings deliberately not fixed

- The mise-registry leg burns ~60s of retry sleeps + a container pull per release while the registry entry is pending — judged a fair trade vs. brittle error-string matching; the design comment already documents it.
- Baseline provenance (nothing stops a laptop `run.sh` promotion from reintroducing environment skew) and the 15% tolerance sitting near the observed ±14–19% ratio noise floor — structural policy calls, flagged in #591's notes.
- Refuted during verification: the claim that an explicit `if:` drops the implicit `success()` guard (GitHub docs: a status function is implied unless one is present), and the claim that `mise activate` pollutes the Verify step's captured stdout (run 107's green mise leg disproves it).

Validated: `go test ./internal/release/... ./cmd/mdsmith-release/...`, `check-release-gates`, `check-release-smoke`, fragment regeneration is byte-stable, and `mdsmith check .` is clean. golangci-lint could not run locally (tools/go.mod needs Go ≥ 1.25.8; environment has 1.25.0) — CI covers it.

https://claude.ai/code/session_0123efoHqbw98eAhPYYjaAkp

---
_Generated by [Claude Code](https://claude.ai/code/session_0123efoHqbw98eAhPYYjaAkp)_

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -870,6 +870,10 @@ jobs:
         # install with `skipped=true` when the upstream registry entry
         # does not exist yet; verifying a binary that was never
         # installed would fail with exit 127, so skip alongside it.
+        # Hard install failures are covered by the implicit success()
+        # in this expression — do not add continue-on-error to the
+        # Install step, or a failed install would fall through to here
+        # with no skipped output and re-fail as a confusing exit 127.
         if: steps.install.outputs.skipped != 'true'
         run: |
           got=$(${{ matrix.run }})
--- a/docs/research/benchmarks/README.md
+++ b/docs/research/benchmarks/README.md
@@ -51,7 +51,7 @@ docs read.
 
 | Corpus  | Files | Source                                            |
 | ------- | ----- | ------------------------------------------------- |
-| repo    | 722   | mdsmith's own tracked Markdown (fixtures dropped) |
+| repo    | 766   | mdsmith's own tracked Markdown (fixtures dropped) |
 | neutral | 234   | Rust Book + Rust Reference `src/` (third-party)   |
 
 ### Environment
@@ -90,7 +90,7 @@ mdsmith-only rules so the work class matches the markdownlint
 tools (see `bench-parity.mdsmith.yml`).
 
 **Repo corpus — 766 Markdown files** (median wall time, lower is
-better; `vs mado` is the median ratio to the fastest tool):
+better; `vs mado` is the ratio to mado's median):
 
 | Tool              | Median  | Min     | vs mado |
 | ----------------- | ------- | ------- | ------- |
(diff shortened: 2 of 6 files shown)


jeduden:
🟢 **Merge Queue** — picked up

**Next:** No action needed — you'll get another comment when CI starts on the batch. [View merge queue run](https://github.com/jeduden/mdsmith/actions/runs/27405839854).

jeduden:

Merged into batch branch [`merge-queue/batch-592-1781254927`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-592-1781254927). [View CI run](https://github.com/jeduden/mdsmith/actions/runs/27405878858).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`ee95624`](https://github.com/jeduden/mdsmith/commit/ee9562492d8033f99caea2650128f3f92edb685c). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/27405878858).

PR #593, opened 2026-06-12
Code-review round 2: pin the soft-skip contract against release.yml, harden gate surfaces, drift-proof benchmark prose

jeduden:
Round 2 of the requested xhigh code review, run over #592's diff after it merged. Nine finder angles plus a sweep produced ~30 raw candidates; after verification they reduce to the fixes below (the headline mechanisms of #592 all held up — the round confirmed the `if:` expression's implicit `success()` guard, the YAML comments sitting outside the `install:` block scalar, byte-stable fragment regeneration, and clean vet/tests).

## Fixes

**Gate hardening and surfaces** (`internal/release/`, `cmd/mdsmith-release/`)
- `TestRepoReleaseWorkflowSoftSkipMarkerInSync` pins the contract against the real `release.yml`: renaming the workflow's `skipped=true` literal without updating `softSkipMarker` would have made the gate silently vacuous — now it's a red test. The same test pins `mise-registry` as best-effort, so promoting it into `RequiredSmokeChannels` (the plan-145 follow-up) forces a conscious edit after its skip path is removed.
- The cmd layer gains the missing end-to-end test: a required channel that soft-skips now provably exits 1 through `run()` dispatch, catching e.g. a yaml-tag regression the unit tests can't see.
- `CheckReleaseSmoke`'s two membership maps fold into one channel→install map (matches the package's idiom).
- The subcommand's usage text, command-list one-liner, and failure hint now name the second violation class — previously an operator hitting it got a Reason string but help text describing only missing-channel coverage. `docs/development/release.md`'s gate-guard bullet gets the same one-phrase update (kept to net-zero lines; the file sits at its 380-line MDS022 budget).

**Prose drift-proofing** (`docs/research/benchmarks/`)
- The parity sentence cited the table's literal `1.1x` — a number `gen_fragments.py` regenerates while prose stands still. Now qualitative ("trails mado narrowly"); the table owns the number.
- The mado growth example's 523→722 figures are explicitly "an earlier pair of snapshots", so they no longer read as the current count two sentences after "766 now".
- `gen_fragments.py`'s new comment trims to the invariant (the file documents invariants, not history) and the mado exit message to one line, matching its other exits.

## Found, deliberately not fixed

- `strings.Contains` is a substring tripwire: a required channel's install script merely *mentioning* `skipped=true` in a shell comment would trip the gate (loudly, at PR time — the failure mode is annoying, not silent), and writing the output via a variable would evade it. Documented on the const; parsing shell would be over-engineering for a five-entry matrix.
- `"mado"` is a magic string in both `gen_fragments.py` and `benchcheck.go`'s `BaselineTool` default. Both fail loudly if mado disappears from the data, so the two-source risk is visible when it bites; cross-language constant plumbing wasn't worth it now.
- The `test(release)` prefix on merged commit `5cbec04` understated its production-code change — merged history, noted for next time.

Validated: `go test ./internal/release/... ./cmd/mdsmith-release/...` (17 smoke-gate tests incl. 3 new), `go vet`, `check-release-gates`, `check-release-smoke`, byte-stable fragment regeneration, `mdsmith check .` clean. golangci-lint needs Go ≥ 1.25.8 (environment has 1.25.0) — CI covers it.

https://claude.ai/code/session_0123efoHqbw98eAhPYYjaAkp

---
_Generated by [Claude Code](https://claude.ai/code/session_0123efoHqbw98eAhPYYjaAkp)_

### DIFF
--- a/cmd/mdsmith-release/main.go
+++ b/cmd/mdsmith-release/main.go
@@ -54,7 +54,7 @@ Commands:
   stamp <version>                 Rewrite tracked manifests to <version>.
   check                           Verify tracked manifests are at the dev sentinel.
   check-release-gates             Verify release-environment gating across .github/workflows.
-  check-release-smoke             Verify release.yml smoke-tests every required channel.
+  check-release-smoke             Verify release.yml smoke-tests every required channel (no soft-skips).
   build-npm <artifacts> <out>     Build npm platform sub-packages.
   build-wheels <artifacts> <out>  Build platform-tagged Python wheels.
   build-flatpak <art> <out>       Stage the .flatpak bundle's manifest + Linux binaries.
@@ -296,7 +296,9 @@ func runCheckReleaseSmoke(root string, args []string) int {
 			"`mdsmith version` reports the tag, so a channel broken at the\n"+
 			"source — like the go.mod replace directive that made v0.40.0\n"+
 			"uninstallable via `go install` — fails the release pipeline\n"+
-			"instead of a user. Used by the release-gate-guard CI job.\n"+
+			"instead of a user. A required channel must also not write the\n"+
+			"best-effort `skipped=true` output that suppresses the shared\n"+
+			"Verify step. Used by the release-gate-guard CI job.\n"+
 			"Exits non-zero on any violation.\n",
 			strings.Join(release.RequiredSmokeChannels, ", "))
 	}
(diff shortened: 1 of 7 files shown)


jeduden:
🟢 **Merge Queue** — picked up

**Next:** No action needed — you'll get another comment when CI starts on the batch. [View merge queue run](https://github.com/jeduden/mdsmith/actions/runs/27408527887).

jeduden:

Merged into batch branch [`merge-queue/batch-593-1781258196`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-593-1781258196). [View CI run](https://github.com/jeduden/mdsmith/actions/runs/27408562684).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`047f1d1`](https://github.com/jeduden/mdsmith/commit/047f1d1431f9e12b4c3d7a8dc4e081be3f7dd21c). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/27408562684).

PR #600, opened 2026-06-13
perf: cut check CPU across YAML, gitignore, glob, and word-count hot paths

jeduden:
## Goal

Drive `mdsmith check` (full default rule set) toward gomarklint's wall time on the two benchmark corpora.

## What changed

Four profiling-driven rounds, each behavior-preserving and verified against the full test suite:

1. **Single-parse YAML safety wrappers** (`internal/yamlutil`). `UnmarshalSafe` / `UnmarshalNodeSafe` parsed every payload twice — once into a `yaml.Node` tree for anchor/alias rejection, then a full `yaml.Unmarshal` re-parse. `parseSafeDocuments` keeps the node from the rejection scan and decodes it directly. This halves YAML parse cost on every front matter and directive body (~19% of repo-corpus CPU before the fix).
2. **Zero-alloc gitignore rel paths + out-of-tree short-circuit** (`internal/gitignore`). `matchRule` called `filepath.Rel` per rule per candidate path (~8% of repo-corpus CPU). `relTo` strips the base prefix without allocation, and paths outside a rule's base (every ancestor-`.gitignore` rule) now return "no match" without paying `Rel`'s Clean allocations.
3. **Cached glob validation + byte-table word counts + lazy token-budget label.** `globpath.Match` re-validated its pattern through `doublestar.Match` on every call; a validity cache plus `MatchUnvalidated` removes the repeated validation walk. `CountWords` gains an ASCII byte-table fast path, mirrored in a new `CountWordsBytes` that token-budget uses directly on `f.Source` instead of copying each whole file into a string; the diagnostic mode label is now built only when a file is actually over budget.
4. **WalkDir-based file walk, stat-free catalog globs, allocation-free `CountSentences`.** `lint.walkDir` moves from `filepath.Walk` (one extra lstat per entry) to `filepath.WalkDir`; catalog glob resolution reuses `GlobWalk`'s `DirEntry` instead of a follow-up `fs.Stat` per match (each avoided stat was an open/fcntl/fcntl/stat/close sequence through the os.Root-backed DirFS); `CountSentences` no longer materialises `[]rune(text)` per call.

## Measured effect (local 4-core, harness-equivalent setup: corpus built per `internal/release/bench.go`, run from the repo root, pinned gomarklint v3.2.3, median of 12 runs)

| Corpus | mdsmith before | mdsmith after | gomarklint | gap before | gap after |
| --- | --- | --- | --- | --- | --- |
| repo (768 files) | 240 ms | 209 ms | ~80 ms | 3.4x | 2.5x |
| neutral (233 files) | 175 ms | 163 ms | ~40 ms | 4.4x | 4.1x |

Total syscalls on the repo corpus drop from ~41.4k to ~37.8k; sys time from ~148 ms to ~95 ms; user+sys CPU from ~720 ms to ~640 ms.

## Honest assessment of the remaining gap

Full parity with gomarklint is not reachable by micro-optimization while the default rule set keeps its current scope. The post-change profile is dominated by work gomarklint does not do at all:

- goldmark AST parse (~17% repo CPU) — gomarklint is line-oriented;
- catalog/include generated-section regeneration and the cross-file front-matter/glob caches (~20% repo CPU, mostly once-per-run);
- the Punkt sentence segmenter behind MDS024 (~27% of neutral CPU) — opt-in by default but enabled by this repo's `.mdsmith.yml`, which both benchmark rows run under; plan 187 already established no faster pure-Go equivalent and a sound skip-gate stronger than the existing `cheapBounds` would amount to re-implementing the annotation pipeline;
- cross-file link integrity, token budgets, and readability scoring.

That structural framing matches the benchmark README: `mdsmith-parity` (same work class as the markdownlint-family tools) already runs in gomarklint/mado's class; the `mdsmith` → `parity` delta is the cross-file and generated-content layer users opt into. The wins here narrow the full-ruleset gap (repo corpus 3.4x → 2.5x) and also speed up the parity row, since the YAML, gitignore, walk, and word-count paths are shared.

Gated benchmarks (`BenchmarkCheckCorpusSmall/Large`), the alloc-budget gate, `golangci-lint`, and `mdsmith check .` all pass.

https://claude.ai/code/session_01D2km9eEecgA2XhgcB6gmLS

---
_Generated by [Claude Code](https://claude.ai/code/session_01D2km9eEecgA2XhgcB6gmLS)_

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -187,4 +187,7 @@ footer: |
 | 2606111049 | ✅     |        | [Harden WASM size test to match production build](plan/2606111049_wasm-size-test-hardening.md)                                          |
 | 2606111050 | ✅     |        | [Guard schema.chmodFile with a mutex like fix.chmodFile](plan/2606111050_schema-chmodfile-mutex.md)                                     |
 | 2606120633 | ✅     |        | [Wire PGO into the release build without a committed profile](plan/2606120633_release-built-pgo-profile.md)                             |
+| 2606130836 | 🔲     | opus   | [Pool the per-file source-read buffer across lintFile passes](plan/2606130836_pool-source-read-buffer.md)                               |
+| 2606130837 | 🔲     | opus   | [Fast-path front-matter field reads for cross-file rules](plan/2606130837_frontmatter-fast-scan.md)                                     |
+| 2606130838 | 🔲     | sonnet | [Memoize linkgraph link and image extraction on the File](plan/2606130838_memoize-link-extraction.md)                                   |
 <?/catalog?>
--- a/internal/gitignore/gitignore.go
+++ b/internal/gitignore/gitignore.go
@@ -261,10 +261,27 @@ func (m *Matcher) DirChainIgnored(dir string, memo map[string]bool) bool {
 	return v
 }
 
+// relTo returns path relative to base, matching filepath.Rel for the
+// inputs the matcher sees (absolute, cleaned paths). The fast path is a
+// zero-allocation prefix strip — filepath.Rel re-cleans both arguments
+// on every call, which dominated the matcher's CPU when every rule of
+// every .gitignore recomputed it per candidate path. Inputs that do not
+// share the base prefix fall back to filepath.Rel for full fidelity.
+func relTo(base, path string) (string, error) {
+	if base == path {
+		return ".", nil
+	}
+	if len(path) > len(base) && path[len(base)] == filepath.Separator &&
+		path[:len(base)] == base {
+		return path[len(base)+1:], nil
+	}
+	return filepath.Rel(base, path)
+}
+
 // matchRule checks whether a single rule matches the given absolute path.
 func matchRule(r ignoreRule, absPath string) bool {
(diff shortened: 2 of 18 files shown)


jeduden:
🟢 **Merge Queue** — picked up

**Next:** No action needed — you'll get another comment when CI starts on the batch. [View merge queue run](https://github.com/jeduden/mdsmith/actions/runs/27466212284).

jeduden:

Merged into batch branch [`merge-queue/batch-600-1781352177`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-600-1781352177). [View CI run](https://github.com/jeduden/mdsmith/actions/runs/27466231096).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`9df4932`](https://github.com/jeduden/mdsmith/commit/9df4932b7512a9b6f3b48a5d4a21225b2780e466). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/27466231096).

PR #647, opened 2026-06-17
perf(engine): unify the two Layer-0 parse-skip gates (plan 2606171400)

jeduden:
## Summary

Implements plan **`2606171400`** — the prerequisite for the parity parse-skip migration. Unifies the engine's two separate parse-skip decisions into one config-aware check.

### The problem

Two gates decided parse-skip independently:
- `layer0SkipEligible` → static `rulelayer.IsLayer0` set (block rules via `BlockChecker`).
- `computeFlatLayer0Active` → config-aware `rule.LineCapable` check (line rules via `ClassifyLines`).

A rule whose layer depends on config — **MDS001 line-length**, line-capable until a per-heading limit is set — was marked AST-forcing by the block gate for *every* config, so it forced the parse even under a config (like parity) where it reads only `f.Lines`.

### The fix

`allEnabledRulesSkipSafe` accepts a rule when it is `rulelayer.IsLayer0` **OR** its configured instance reports `rule.LineCapable` (`ruleConfiguredLineCapable`, now shared with `computeFlatLayer0Active` — one eligibility helper). The skip File is already `NewFileFlatPooled`, which carries both the `ClassifyLines` projection and the block-span scan, so a line-capable rule serves the nil-AST path with no further change. The unknown-rule default stays AST-forcing.

### Tests (TDD red→green)

- `TestLayer0Gate_LineCapableRuleSkipsParse` — line-length now skips the parse (was the failing test).
- `TestLayer0Gate_LineCapableDiagnosticsMatchFullParse` — skip vs full-parse identical.
- `TestLayer0Gate_LineCapableCorpusEquivalence` — **corpus-wide** byte-identical guard with line-length on the skip path.
- `TestLayer0Gate_ParsesWhenAnASTRuleIsEnabled` updated to use line-length **with** `heading-max`, so it still exercises a genuinely AST-forcing rule.

## Scope

This does **not** move the benchmark on its own — the full parity set still has unmigrated AST-forcing rules that block the all-or-nothing gate. It's the foundation that lets config-dependent line rules count once their batches land. Full `go test ./...` passes; `gofmt`/`vet` clean (golangci-lint runs in CI — the sandbox Go is below the dev-tool floor).

https://claude.ai/code/session_01LmDJshZDHCDwWzv5jCnRgt

---
_Generated by [Claude Code](https://claude.ai/code/session_01LmDJshZDHCDwWzv5jCnRgt)_

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -206,7 +206,7 @@ footer: |
 | 2606162214 | ✅     | sonnet | [Add named unit tests for unexported helpers in new lazy-parse packages](plan/2606162214_arch-fix-missing-helper-tests.md)              |
 | 2606171136 | 🔳     | opus   | [Parity perf: per-worker configured-rule cache + the Layer-0 path to gomarklint](plan/2606171136_parity-perf-configured-rule-cache.md)  |
 | 2606171258 | 🔳     | opus   | [Parity Layer-0 parse-skip: migrate the AST-forcing parity rules](plan/2606171258_parity-layer0-parse-skip.md)                          |
-| 2606171400 | 🔲     | opus   | [Parity parse-skip: unify the two Layer-0 gate mechanisms](plan/2606171400_parity-gate-unification.md)                                  |
+| 2606171400 | ✅     | opus   | [Parity parse-skip: unify the two Layer-0 gate mechanisms](plan/2606171400_parity-gate-unification.md)                                  |
 | 2606171401 | 🔲     | sonnet | [Parity parse-skip: migrate the Layer-0 heading and front-matter rules](plan/2606171401_parity-layer0-heading-rules.md)                 |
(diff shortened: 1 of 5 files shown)


jeduden:
🟢 **Merge Queue** — picked up

**Next:** No action needed — you'll get another comment when CI starts on the batch. [View merge queue run](https://github.com/jeduden/mdsmith/actions/runs/27697495067).

jeduden:

Merged into batch branch [`merge-queue/batch-647-1781707554`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-647-1781707554). [View CI run](https://github.com/jeduden/mdsmith/actions/runs/27697558632).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`2cb73df`](https://github.com/jeduden/mdsmith/commit/2cb73dfa681ef8bfe80304252a59b54ba713a4a6). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/27697558632).

PR #664, opened 2026-06-19
Parity Layer-0 parse-skip: admit nil-AST-safe B-prose-only rules (closes plan 2606171258)

jeduden:
## Summary

Plan [`2606171258`](plan/2606171258_parity-layer0-parse-skip.md) was completed on `main` with a `blockSkipSafe` allow-list that promotes the single rule **MDS066** to Layer 0. This PR replaces that per-rule special case with a **category rule**: `rulelayer.nilASTBackable` promotes the whole nil-AST-safe `B-prose-only` category to Layer 0, gated on the manifest's `nil_ast_safe` signal.

Per maintainer direction, this is the deeper "generalize the mechanism over a special case" approach.

## Why it's sound

`B-prose-only` already implies the audit probe saw **no** nil-AST divergence on the unperturbed fixture (a divergent rule lands in `hybrid`). The only withheld signal is code-sensitivity — and the code guard that once made that a disqualifier is gone, with the `ClassifyLines` and Layer-1 inline projections proven byte-identical to the AST on code content. `astProjectionConsumers` stays as the veto for any rule that reads an unbacked AST projection.

This admits all four nil-AST-safe `B-prose-only` rules (MDS041, MDS050, MDS052, MDS066), not just MDS066. Parity enables only MDS066; the other three are opt-in.

## Soundness guards

- `TestBProseOnlyRulesAreNilASTSafe` — every live `B-prose-only` manifest entry is `nil_ast_safe` and resolves to Layer 0 (the category-promotion invariant).
- `TestNilASTBackable` — dedicated unit test for the promotion predicate (all arms).
- `TestLayer0Gate_ParityCorpusDiagnosticsEquivalence` — parse-skip vs full-parse diagnostics match across the whole corpus with the **full** parity set enabled (line-length + the B-prose-only rules), not just the static Layer 0 subset.
- Each rule's existing `TestCheck_NilASTMatchesAST` checks its own nil-AST path.

## Notes

- Updates the `propernames` reachability comment: MDS050's check-code/check-html reparse branch is now reachable in production (and stays byte-identical).
- `go test ./...` green; `mdsmith check .` clean (500 files).

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -205,7 +205,7 @@ footer: |
 | 2606162213 | ✅     | sonnet | [Document internal/checker and internal/rulelayer; remove engine shim](plan/2606162213_arch-fix-new-pkg-docs.md)                        |
 | 2606162214 | ✅     | sonnet | [Add named unit tests for unexported helpers in new lazy-parse packages](plan/2606162214_arch-fix-missing-helper-tests.md)              |
 | 2606171136 | ✅     | opus   | [Parity perf: per-worker configured-rule cache + the Layer-0 path to gomarklint](plan/2606171136_parity-perf-configured-rule-cache.md)  |
-| 2606171258 | 🔳     | opus   | [Parity Layer-0 parse-skip: migrate the AST-forcing parity rules](plan/2606171258_parity-layer0-parse-skip.md)                          |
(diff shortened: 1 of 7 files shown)


jeduden:
Closing as redundant. Plan `2606171258` was already completed on `main` (the `blockSkipSafe` override promoting MDS066 to Layer 0, with `internal/engine/parity_layer0_test.go` covering all three acceptance criteria). This branch was an independent, competing implementation of the same fix — the merged work supersedes it, so there's nothing left to land here.

---

PR #673, opened 2026-06-20
perf(parity): make the Layer-0 parse-skip correct for list files (and measure why it's still a wash on benchmark 2)

jeduden:
## Goal & honest result

Target: make `mdsmith check -c parity` as fast as gomarklint on benchmark
2 (the 234-file Rust neutral corpus). This PR makes the parse-skip path
**correct** for list-bearing files (it wasn't) and **measures that the
parse-skip is a wash on this corpus** — it does not make parity faster.
The result is recorded honestly rather than shipped as a false win; the
master toggle stays default-off.

This PR's conclusion was that beating gomarklint outright "needs a
rule/overhead trim or **a smaller parity rule set, a product decision**."
The follow-up commits act on that decision (see the last two sections).

## What was broken, and fixed

Three real defects on the Layer-0 (nil-AST) path, each independently
gated:

1. **listscan dropped CommonMark's paragraph-interruption rule at the
   document root.** An ordered marker like `2026.` mid-paragraph started
   a spurious list where goldmark reads lazy paragraph text. Fixed by
   tracking top-level paragraph state. Adds a whole-corpus
   listscan/goldmark equivalence test; corpus divergence 8 files → 2
   (both HTML-comment cases, now a documented listscan limitation).
2. **The list rules' nil-AST paths were dead code in the engine.** MDS014/
   016/045/046/061 are `NodeChecker`s with a working `checkLayer0` branch,
   but the checker dispatch routes a `NodeChecker` only through the AST
   walk — on a nil-AST File it dropped them, so `Check`/`checkLayer0`
   never ran. A forced-skip probe showed 102 missing diagnostics, all
   from these five rules. Fixed with a `rule.LinesChecker` capability
   (mirroring `rule.InlineChecker`) that routes such a rule to its own
   `Check` on a skipped File. Drops the gate's list guard;
   `TestLayer0Gate_CorpusDiagnosticsEquivalence` now engages on list
   files and stays byte-identical.

## Why it's still a wash (measured, 4-core box, min of many runs)

| Run                              | wall   | vs gomarklint |
| -------------------------------- | ------ | ------------- |
| gomarklint                       | ~32 ms | 1.0x          |
| parity, parse-skip off (shipped) | ~84 ms | ~2.6x         |
| parity, parse-skip on            | ~86 ms | ~2.6x         |

- **Eligibility:** the gate disqualifies any file with a `>` (the
  block-quote guard). Rust prose is full of block quotes *and* `>` in
  code (`Vec`, `->`), so only ~26 of 234 files clear it — and the big
  quote-bearing chapters that dominate wall time still parse.
- **Cost:** a skipped file runs three line passes instead of the parse
  (`ClassifyLines` + `scanLayer0` + `listscan.Parse`), which lands in the
  same class as goldmark's single parse. The saving on the eligible files
  is within noise.

Full write-up:
`docs/research/benchmarks/parity-parse-skip-findings.md`.

## Why default-off

The path is correct (the equivalence gates hold it byte-identical), but
correct ≠ faster here, and enabling the gate adds a per-file `>`-scan to
every run. It stays the opt-in `MDSMITH_LAYER0_SKIP=1` seam until the two
levers that would make it a real win land: **descend into block quotes**
(so quote-bearing files become eligible) and **fuse the projections**
(so the skip is cheaper than the parse).

## Follow-up 1: per-linter parity conventions (the "smaller rule set")

The generic `parity` convention is **removed** and replaced by four
per-linter conventions, each running the rule set its peer enables by
default. Each set is derived from the per-rule peer mappings in the
coverage matrix (the README front matter) and verified against it by
`TestParityConventionsMatchCoverageMatrix` in `internal/integration`, so
the conventions can never drift from the matrix.

## Follow-up 2: MDS027 made explicit, partial covers excluded

The peers' `link-fragments` / MD051 rules resolve only **same-file**
`#anchor` links; mdsmith's `cross-file-reference-integrity` (MDS027) also
walks the workspace for **cross-file** links. The matrix claimed a full
match — it's really partial. So:

- The gomarklint/rumdl/markdownlint link-fragments mappings for MDS027 are
  marked `partial: true` (the coverage matrix and the MDS027 peer-link
  block now render "(partial)").
- Parity now runs a peer's rule only on a **full** cover (`default &&
  !partial`), so it never runs the heavier mdsmith rule on a peer's behalf.

Resulting sets and parse-skip status:

| Convention            | Rules | Parse-skip-safe?         |
| --------------------- | ----- | ------------------------ |
| `gomarklint-parity`   | 20    | **Yes** (now drops MDS027) |
| `mado-parity`         | 27    | **Yes**                  |
| `rumdl-parity`        | 41    | No (MDS020 required-headings, a full cover) |
| `markdownlint-parity` | 41    | No (MDS020)              |

**This achieves the gomarklint goal:** `gomarklint-parity` is now fully
parse-skip-safe, so mdsmith can follow gomarklint onto the no-parse path
— what remains is per-rule and fixed overhead, not the goldmark parse.
The trade-off is that it then runs no anchor check, where gomarklint runs
a same-file one. Plan `2606210840_same-file-anchor-resolution-rule.md`
tracks the separate effort: a parse-skip-safe same-file-anchors rule that
restores a like-for-like anchor check and turns the matrix mapping into an
honest full cover.

The benchmark page's `mdsmith-parity` column and the PGO profile select
`mado-parity` (the mid-size representative peer); the four
`bench-<linter>-parity.mdsmith.yml` profiles select their matching
convention. **The committed timing tables are the pre-split measurements
and have not been re-run** (no peer binaries in this environment); a fresh
`mdsmith-release bench` is needed to publish true per-linter columns.

https://claude.ai/code/session_01V1rMBt2mnTcg4f3HrArdF8


jeduden:
🟢 **Merge Queue** — picked up

**Next:** No action needed — you'll get another comment when CI starts on the batch. [View merge queue run](https://github.com/jeduden/mdsmith/actions/runs/27910581338).

jeduden:

Merged into batch branch [`merge-queue/batch-673-1782059563`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-673-1782059563). [View CI run](https://github.com/jeduden/mdsmith/actions/runs/27910600146).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`e701b94`](https://github.com/jeduden/mdsmith/commit/e701b943759e53c8073783211afbadb2b9712d81). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/27910600146).

PR #688, opened 2026-06-23
perf: apply high-performance-go guidelines across 5 hot paths (3-round review)

jeduden:
## Summary

Performance audit against `docs/development/high-performance-go.md`, fixing 5 violations across core packages. Three rounds of `code-review --fix` at high severity were applied before merging.

### Fixes (commit `7c9dfc7`)

| Location | Violation | Fix |
|---|---|---|
| `internal/export/export.go` | `map[int]bool` sets for line tracking | `map[int]struct{}` — zero-size value, same semantics |
| `internal/index/locate.go` `enclosingListKey` | `FindStringSubmatch(string(lines[i]))` — one `string()` copy per scanned line | `FindSubmatch(lines[i])` — regex operates directly on `[]byte` |
| `internal/lsp/rename.go` / `internal/rename/rename.go` | `ValidRefDefBodyLines` returned `map[int]bool`; lookup used `map[bool]` zero-value | Return `map[int]struct{}`; lookups use `_, ok` idiom |
| `internal/secreview/render.go` `locStr` | `fmt.Sprintf(":%d", n)` — reflection overhead for integer formatting | `":" + strconv.Itoa(n)` — ~3× faster, no reflection |
| `internal/rules/concisenessscoring/rule.go` | `message += fmt.Sprintf(...)` — intermediate string allocation on verbose path | Single `fmt.Sprintf` via if/else — no extra allocation |

### Code-review round 1 fixes (commit `917eafa`)

- `rule_test.go` `TestCheck_MessageNoConcatenationWhenExamplesPresent`: replaced vacuous `assert.NotEmpty` with `assert.Contains(msg, "e.g.,")` — the original guard could never fail
- `locate_test.go`: renamed `TestEnclosingListKey_NoStringAllocPerLine` → `TestEnclosingListKey_FindsParentKey` (old name implied allocation testing the test didn't do)
- `rule_test.go`: added `TestCheck_NoCuesMessage` to cover the `examples == ""` branch, which was missing coverage (Codecov patch gate was failing at 82%)

### Code-review round 2 fixes (commit `12f6a79`)

- `locate.go` `enclosingListKey`: removed redundant `bytes.TrimSpace(m[2])` — piArgRE's `\s*$` suffix structurally guarantees group 2 has no trailing whitespace; replaced with `len(m[2]) == 0`
- `rule_test.go` `TestCheck_NoCuesMessage`: replaced hardcoded `MinScore: 0.50` (fragile against model drift) with a runtime scorer probe — calls `NewScorer()`, asserts `len(Cues) == 0`, and sets `MinScore = scored.Conciseness + 0.10`

### Code-review round 3 fixes (commit `3a45453`)

- `rule_test.go` `TestCheck_MessageNoConcatenationWhenExamplesPresent`: removed conditional guard `if strings.Contains(msg, "reduce verbose cues")` — the inner `assert.Contains(msg, "e.g.,")` was dead code in the exact regression it was meant to catch (if implementation drops cue text, the outer `if` is false and the inner assert never runs); `verboseParagraph()` always produces cues so both asserts are now unconditional
- `rule.go`: unified the two `fmt.Sprintf` call sites (sharing the same base format and three repeated arguments) into a single call with a conditional `cuesSuffix` string — base format string now lives in one place

## Test plan

- [x] All tests pass: `go test ./...`
- [x] Codecov patch coverage ≥ 98.62% (green on all 4 commits)
- [x] Three rounds of `code-review --fix` at high severity — no surviving CONFIRMED findings after round 3
- [x] `go vet ./...` clean

https://claude.ai/code/session_015QYub2iGP6CeeEFSfjV3Wv

### DIFF
--- a/internal/export/export.go
+++ b/internal/export/export.go
@@ -244,18 +244,18 @@ func inGeneratedRange(line int, ranges []lint.LineRange) bool {
 // because such PIs sit inside a pair's body range and are skipped
 // here.
 func stripDirectives(f *lint.File, directives []directiveStrip) []byte {
-	stripLines := map[int]bool{}
-	bodyLines := map[int]bool{}
+	stripLines := map[int]struct{}{}
+	bodyLines := map[int]struct{}{}
 
 	for _, d := range directives {
 		pairs, _ := gensection.FindMarkerPairs(f, d.name, d.ruleID, d.ruleName)
 		for _, p := range pairs {
 			for line := p.StartLine; line < p.ContentFrom; line++ {
-				stripLines[line] = true
+				stripLines[line] = struct{}{}
 			}
-			stripLines[p.EndLine] = true
+			stripLines[p.EndLine] = struct{}{}
 			for line := p.ContentFrom; line <= p.ContentTo; line++ {
-				bodyLines[line] = true
+				bodyLines[line] = struct{}{}
 			}
 		}
 	}
@@ -276,7 +276,7 @@ func stripDirectives(f *lint.File, directives []directiveStrip) []byte {
 			continue
 		}
 		for line := startLine; line <= endLine; line++ {
-			stripLines[line] = true
+			stripLines[line] = struct{}{}
 		}
 	}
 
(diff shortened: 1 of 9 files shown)


jeduden:
🟢 **Merge Queue** — picked up

**Next:** No action needed — you'll get another comment when CI starts on the batch. [View merge queue run](https://github.com/jeduden/mdsmith/actions/runs/28059134933).

jeduden:

Merged into batch branch [`merge-queue/batch-688-1782251381`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-688-1782251381). [View CI run](https://github.com/jeduden/mdsmith/actions/runs/28059454470).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`09f22d3`](https://github.com/jeduden/mdsmith/commit/09f22d3a59ea9cc07911df1db2d462da7d5fccb1). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/28059454470).

PR #725, opened 2026-07-07
perf: fix top 5 high-performance-go.md violations in the rule hot path

jeduden:
## Summary

An audit against `docs/development/high-performance-go.md` scanned `internal/rules`, `internal/lint`, `internal/mdtext`, `pkg/markdown`, `internal/engine`, and `internal/config` for the four anti-pattern classes the doc calls out (un-presized allocations, string/bytes conversions, structural issues, and skipped/redundant work). This PR fixes the top 5 findings, ranked by how hot the call site is (default-on rule × every workspace file > opt-in rule / gated path).

1. **`astutil.CollectSectionHeadings` was not memoized**, unlike its sibling `CollectSectionParagraphs`. MDS057 (`required-text-patterns`) and MDS058 (`required-mentions`) both call it from `Check`, so enabling both re-walked the same AST twice per file. Now routed through `lint.File.MemoFile`, matching the existing `buildSectionParagraphs` pattern.
2. **`requiredstructure.extractHeadings` was not memoized** despite 4 call sites in `Check`/`Fix` (`checkSingleFileSchemaFromData`, the per-source loop in `checkComposedSources` via `bodySyncDiagnostics`, `checkSync`, `fixBodySyncIn`). A schema with N `extends` sources re-walked the AST N+ times per `Check`. Now memoized the same way.
3. **`headingincrement.Check`/`checkNilAST`** (MDS003, default-on, runs on every file) unconditionally allocated `make([]lint.Diagnostic, 0, 4)` even when no heading violates. Switched to a nil-starting slice per the doc's "Return nil, not `[]T{}`" convention.
4. **`tableformat.structureDiagnostics`** (MDS025, default-on for files containing `|`) pre-sized `diags` to `len(tables)`, allocating even when every table is well-formed. Switched to a nil-starting slice (with a `//nolint:prealloc` matching the existing precedent in `paragraphstructure/rule.go`).
5. **`noreferencestyle.scanFootnoteReferences`/`scanFootnoteDefinitions`** grew `out` from `nil` via `append` even though `FindAllSubmatchIndex`'s match count is a known upper bound. Pre-sized with `make([]footnoteOccurrence, 0, len(matches))`.

Four parallel scan agents also checked regex-compile-in-hot-path, string/byte conversions, defer-in-loop, `time.Now()`/`context.Background()` misuse, `os.ReadFile` on the hot path, and `[]*Foo` vs `[]Foo` — the codebase already applies those patterns correctly (documented in-line at several call sites), so those categories produced no additional actionable findings beyond the 5 above.

## Test plan

- [x] Each fix has a dedicated red/green unit test: alloc-count regression tests (`testing.AllocsPerRun`) for #5, `assert.Nil` convention tests for #3/#4, and memoization identity tests (`assert.Same`) for #1/#2 — each confirmed failing against the pre-fix code, then passing after the fix (see individual commits).
- [x] `go test ./...` — all 151 packages pass
- [x] `go build ./...` and `go vet ./...` clean
- [x] `go run ./cmd/mdsmith check .` — 0 failures

https://claude.ai/code/session_01BTLNzEVChZtqSc7m7ph5Ah

---
_Generated by [Claude Code](https://claude.ai/code/session_01BTLNzEVChZtqSc7m7ph5Ah)_

### DIFF
--- a/internal/rules/astutil/astutil.go
+++ b/internal/rules/astutil/astutil.go
@@ -68,7 +68,19 @@ func (p SectionParagraph) ExtractText(source []byte) string {
 // CollectSectionHeadings returns every heading in the document
 // ordered by source line. Used by content rules (MDS057, MDS058)
 // that need to walk heading-bounded sections.
+//
+// Memoized per File via lint.File.MemoFile, mirroring
+// CollectSectionParagraphs: MDS057 and MDS058 both enabled would
+// otherwise re-walk the same AST for the same result.
 func CollectSectionHeadings(f *lint.File) []SectionHeading {
+	return f.MemoFile("astutil.sectionHeadings", buildSectionHeadings).([]SectionHeading)
+}
+
+// buildSectionHeadings is the MemoFile-style builder for the
+// section-headings memo. Defined at package scope so the value passed
+// to MemoFile is a plain function pointer, matching
+// buildSectionParagraphs.
+func buildSectionHeadings(f *lint.File) any {
 	var out []SectionHeading
 	_ = ast.Walk(f.AST, func(n ast.Node, entering bool) (ast.WalkStatus, error) {
 		if !entering {
--- a/internal/rules/astutil/astutil_test.go
+++ b/internal/rules/astutil/astutil_test.go
@@ -523,6 +523,21 @@ func TestCollectSectionHeadings_NoHeadings(t *testing.T) {
 	assert.Empty(t, CollectSectionHeadings(f))
 }
 
+// TestCollectSectionHeadings_Memoized pins that repeated calls share
+// the same backing slice — MDS057 and MDS058 both enabled should pay
+// the AST walk once, not twice, mirroring
+// TestCollectSectionParagraphsWithText_Memoized below.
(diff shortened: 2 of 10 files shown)


jeduden:
🟢 **Merge Queue** — picked up

**Next:** No action needed — you'll get another comment when CI starts on the batch. [View merge queue run](https://github.com/jeduden/mdsmith/actions/runs/29099653301).

jeduden:

Merged into batch branch [`merge-queue/batch-725-1783693464`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-725-1783693464) alongside #729. [View CI run](https://github.com/jeduden/mdsmith/actions/runs/29099729735).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`24164df`](https://github.com/jeduden/mdsmith/commit/24164df1e1ec9161c5edac4de14a3857a19b42e2). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/29099729735).

PR #729, opened 2026-07-09
perf: fix top 5 high-performance-go.md violations found by audit

jeduden:
## Summary

Audited the codebase against [`docs/development/high-performance-go.md`](docs/development/high-performance-go.md) using a team of scan agents covering rule packages, core packages, static-analysis tooling, and existing benchmark budgets. Selected and fixed the top 5 real, safely-fixable violations, each with red/green TDD and measured evidence (`testing.AllocsPerRun`, `benchstat`).

**Not included:** MDS025 (table-format) is the single biggest raw offender (52 allocs/op vs. a 10-alloc ceiling), but its fix is an already-tracked, deliberately-deferred architectural refactor (plan 195 task 3 — a "single-table-walk" rewrite merging two independent table parses). Attempting that rewrite of ~2,300 lines of heavily-tested table-parsing code in one pass was judged too risky; it's called out here for visibility rather than reworked.

## Fixes

1. **`orderedlistnumbering` (MDS046): `fmt.Sprintf` → `strconv` in the per-item hot loop.** Fires once per violating ordered-list item. `benchstat`: **-42% CPU, -9% bytes/op** (p=0.000), allocs/op unchanged.

2. **Struct field layout — `internal/index` (`Edge`, `Symbol`).** Both interleaved scalar `int` fields between `string` fields, forcing Go's GC ptrdata span to cover bytes it doesn't need to scan. Reordered so pointer-containing fields precede scalars, per the doc's "Struct layout" section. These types are allocated once per link/symbol across the whole workspace index build and every LSP `Update`.

3. **Struct field layout — `mdtext.TOCItem`, `schema.DocHeading`, `schema.ScopeMatch`.** Same pattern, three more hot-path types (TOC fix pass, schema validation).

4. **`fmt.Sprintf` → `strconv` — `tokenbudget`, `maxfilelength`, `firstlineheading`.** Interesting finding: commit `e05a866` previously reverted this exact swap for two of these rules, citing a measured allocation *regression* on an earlier Go toolchain. Re-measured on today's toolchain (go1.25.11), `strconv.Itoa` now caches formatted results for 0-99, so the previously-measured gap no longer reproduces — verified directly before and after each change. `benchstat` shows a real, allocation-neutral CPU win across all three (-19% to -31% CPU, p=0.000).

5. **Enable `golangci-lint`'s `perfsprint` and `gocritic`(performance) linters**, exactly as the doc's Tooling section recommends but `.golangci.yml` didn't yet do. Both surfaced far larger surfaces than expected once uncapped (`gocritic`'s `enabled-tags` turned out to be *additive*, not restrictive, and golangci-lint's default issue caps were silently truncating `run` output). Fixed the 26 safe one-line findings (`stringXbytes`, `appendCombine`, `preferStringWriter`, `equalFold`) and the 10 genuine `fmt.Sprintf`→concat findings; scoped out ~180 cold-path `fmt.Errorf`→`errors.New` findings (via `perfsprint.error-format: false`) and 263 `hugeParam`/`rangeValCopy` findings that need function-signature changes across many files (via a documented `exclude-rules` entry) as separately-reviewable follow-up debt, mirroring this repo's existing "grandfather" convention (see `allocBudgetGrandfathered` in `internal/integration/alloc_budget_test.go`).

## Test plan

- [x] `go build ./...`
- [x] `go test ./...` (all packages pass)
- [x] `go vet ./...`
- [x] `gofmt -l .` clean
- [x] `go run ./cmd/mdsmith check .` — 532 files checked, 0 failures
- [x] Each fix has a dedicated red/green test and, where a CPU (not allocation) win was the claim, a `benchstat`-verified benchmark

---
_Generated by [Claude Code](https://claude.ai/code/session_01WNJitFyE4usUhymZTegXAG)_

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -9,20 +9,6 @@ linters:
   exclusions:
     paths:
       - pkg/goldmark
-    rules:
-      # gocritic's hugeParam/rangeValCopy (both performance-tagged)
-      # flag ~260 existing by-value struct params and for-range loops
-      # across cmd/mdsmith, internal/build, internal/config, and
-      # internal/schema — genuine wins, but each fix touches a
-      # function signature or call sites, and some types are used
-      # across package boundaries. That is a much larger, separately
-      # reviewable pass than a single perf-audit PR; tracked as
-      # follow-up debt per docs/development/high-performance-go.md
-      # Tooling rather than grandfathered silently.
-      - linters: [gocritic]
-        text: "^hugeParam:"
-      - linters: [gocritic]
-        text: "^rangeValCopy:"
   enable:
     - errcheck
     - funlen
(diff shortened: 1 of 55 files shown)


jeduden:
🟢 **Merge Queue** — picked up

**Next:** No action needed — you'll get another comment when CI starts on the batch. [View merge queue run](https://github.com/jeduden/mdsmith/actions/runs/29099653301).

jeduden:

Merged into batch branch [`merge-queue/batch-725-1783693464`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-725-1783693464) alongside #725. [View CI run](https://github.com/jeduden/mdsmith/actions/runs/29099729735).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`24164df`](https://github.com/jeduden/mdsmith/commit/24164df1e1ec9161c5edac4de14a3857a19b42e2). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/29099729735).

PR #732, opened 2026-07-11
perf: fix top 5 high-performance-go.md violations

jeduden:
## Summary

An audit against `docs/development/high-performance-go.md` (dispatched as parallel research agents scanning for allocation, string/bytes, struct-layout, and skip-work anti-patterns) surfaced several violations of the project's performance guidelines. This PR fixes 4 of them, ranked by hotness and impact, each with red/green TDD and a benchmark or deterministic test pinning the win. (A 5th finding — MDS037 `duplicated-content`'s corpus-index caching — was dropped from this PR after rebasing onto `main`: PR #723 independently fixed the same hotspot with a different design, `RunCache.DuplicateParagraphs`, which merged first.)

1. **`internal/fix` (mdsmith fix pipeline)** — every enabled `Configurable` rule was re-cloned and re-`ApplySettings`'d up to 3× per file (once in `fixableRules`, once in each of the pre-/post-fix `CheckRules` calls), even though `internal/engine` already memoizes this exact work per config signature for `mdsmith check`. Added `Fixer.effCache`/`confCache` mirroring `engine.Runner`'s caches, so a fix run over many files sharing one config configures each rule once for the whole run. `BenchmarkFixCorpus` (60 files): ~9.3k → ~8.4k allocs/op.

2. **`pkg/markdown/flavor.LineCol`** — counted newlines with a hand-rolled byte loop over `source[0:offset]`, called once per MDS034 finding — O(n) per finding instead of O(n) total. Swapped the internal implementation (same pinned public signature) to `bytes.Count`/`bytes.LastIndexByte` (SIMD-accelerated). `BenchmarkLineColLateOffset`: ~45.1µs/op → ~2.4µs/op (-94.7%, benchstat p=0.008).

3. **MDS028 `token-budget`** (default-enabled) — ran a full-file word/regex scan on every file regardless of size, even though most files sit well under the default 8000-token budget. Added `definitelyUnderBudget`, a cheap `len(source)`-only upper-bound check that lets `Check` skip the real scan entirely for the common case. `TestCheck_AllocBudget_UnderBudget` pins the fast path at 0 allocs/op.

[...]

PR #731, opened 2026-07-10
perf: fix top 5 high-performance-go.md violations from audit

jeduden:
## Summary

An agent-team audit against `docs/development/high-performance-go.md` scanned the hot per-file `check`/`fix` path (allocation patterns, strings/bytes, concurrency/skip-work, and struct layout) across `internal/rules/`, `internal/lint/`, `pkg/markdown/`, and `internal/mdtext/`. This PR fixes the top 5 ranked findings, each with a red/green benchmark or allocation/layout regression test added before the fix, then survives three independent xhigh-severity code-review passes (8 finder angles each, findings verified before fixing).

One additional candidate (removing a `defer`-wrapped `sync.Pool` release from `astutil.HeadingText`/`HeadingTextBase`) was investigated and measured with `benchstat` (n=15, `-count=15`) but showed no significant effect (`p=0.467`) at that call site, so it was dropped in favor of a better-evidenced fix, per the guideline's own "decide with `benchstat`, not eyeballs" standard.

## The 5 fixes

1. **`internal/rules/linkstyle`** — `checkExtension` called `strings.ToLower` on every link/image target in every checked file (MDS022, a default rule — one of the hottest per-node paths in the rule set). Switched to `strings.EqualFold`, which compares without allocating. *(docs/development/high-performance-go.md#strings-and-bytes)*

2. **`internal/rules/duplicatedcontent`** — `isMarkdownPath` lower-cased the whole path on every sibling file while indexing the corpus (O(files) per run). Switched to an extension-only case-insensitive check. *(docs/development/high-performance-go.md#strings-and-bytes)*

3. **`internal/rules/requiredstructure`** — `buildFieldPattern` called `regexp.MustCompile` on every `{field}`-bearing body-sync line, rebuilding an identical NFA whenever a schema's body repeated the same template line. Added a call-scoped cache keyed by body text. *(docs/development/high-performance-go.md#allocations — "Compile regexes at package scope")*

4. **`internal/lint`** — `BlockSpan` interleaved a 1-byte `Kind` field before three 8-byte ints and a trailing 1-byte `Closed` field, padding the struct to 40 bytes. `BlockSpan` is built once per block by `Layer0` and read by `rule.WalkBlocks` across every default rule on the parse-skip path (MDS002, MDS010, MDS011, MDS013, MDS015, MDS031, MDS065, MDS066). Reordering large-to-small packs it into 32 bytes. *(docs/development/high-performance-go.md#struct-layout)*

[...]

PR #723, opened 2026-07-06
perf: fix top 5 high-performance-go.md violations

jeduden:
## Summary

An audit against [`docs/development/high-performance-go.md`](docs/development/high-performance-go.md) (using a team of exploration agents covering allocations/regexp, data-structure/struct-layout, and memoization/concurrency anti-patterns) surfaced a ranked list of violations. Several strong candidates were already tracked by existing plans with deliberate grandfathered debt (`plan/195` for `tableformat`'s double table-parse, `plan/2606130838` for `crossfilereferenceintegrity`'s 12-alloc budget, `plan/189` for the AST-walk migration, which explicitly excludes MDS003/MDS005/MDS059/MDS065 as stateful passes) — those were left alone to avoid duplicating or contradicting in-flight work.

The five fixes below are fresh, safe, and each independently testable:

1. **`internal/rules/astutil`**: memoize `CollectSectionHeadings` via `f.MemoFile`, mirroring the already-memoized `CollectSectionParagraphs` in the same file. MDS057 and MDS058 each re-walked the AST from scratch; now the walk is shared.
2. **`internal/rules/build`**: `warnUnknownParams`'s per-directive `known` set and the package-level `reservedDeviceNames` set used `map[string]bool` for membership-only checks. Converted to `map[string]struct{}` per the "zero-byte value type" guidance, reading through the existing `internal/setutil.Contains` helper.
3. **`pkg/markdown/flavor`**: replaced the `map[Flavor]map[Feature]bool` support table (consulted once per candidate AST node during flavor detection) with a flat `[flavorCount][featureCount]bool` array, removing two map indirections per call. `Supports` keeps its "false for an out-of-range value" contract (this is a public package) via an explicit bounds check, and two new regression tests pin `flavorCount`/`featureCount` against the real enum surface so a future constant added without updating them fails a test instead of silently losing its row.
4. **`pkg/goldmark/parser` and `pkg/goldmark/ast`**: reordered `Delimiter`, `linkLabelState`, `fenceData`, and `AutoLink` so pointer/interface fields precede scalar fields. GC ptrdata spans through the last pointer field, so the previous ordering forced the GC to scan scalar tails on every one of these hot per-token structs (allocated per emphasis-run, per bracket candidate, per fenced-code-block, and per autolink in every file's inline parse). The reflection-based layout tests share one helper (`pkg/goldmark/internal/fieldorder`) instead of duplicating it per package.
[...]

PR #736, opened 2026-07-12
test(architecture): add missing tests for public rule/directive contract methods

jeduden:
## Summary

Scheduled SOLID-architecture audit of the 66 production Go files touched since the last checkpoint (`528ce4c..834b560`) — every modified `internal/rules/*` package, `cmd/mdsmith`, `cmd/mdsmith-release`, `cue/cuelite`, a dozen core `internal/*` packages, and the `pkg/goldmark`/`pkg/markdown/flavor`/`pkg/mdsmith` public-API layer.

Full findings are recorded under `## Audit 2026-07-12` in [docs/development/architecture-audit.md](docs/development/architecture-audit.md). This PR fixes the highest-priority (blocker-severity) finding that was safe to fix without a behavior change:

- `internal/rules/catalog/rule.go`, `internal/rules/build/rule.go`, and `internal/rules/include/rule.go` implement `gensection.Directive` — a generated-markers contract surface. Their `RuleID`/`RuleName`/`Validate`/`Generate` methods had no dedicated unit test.
- `internal/rules/listindent/rule.go`'s `rule.Rule` contract methods `ID()`/`Name()` had the same gap.
- `internal/index.File()` — the LSP/rename/deps query surface's primary lookup — had no dedicated test of its hit/miss, path-normalization, or copy-not-alias behavior.

Per the project's test-pyramid rule (`docs/development/architecture/tests.md`), a missing unit test is `blocker`-severity when the function sits on a public surface. All five gaps qualify. No production code changed — this is test-only.

## Other findings

- **1 blocker not fixed here** — `internal/linkgraph/wikilinks.go`'s `NewWikilinkIndex`/`ResolveWikiLink` call `fs.WalkDir`, contradicting go.md's "Pure (no file reads, no workspace walks)" claim for the package. This needs an actual design decision (move the walk into `internal/index`, or rescope the doc) rather than a mechanical fix, so it's filed as [plan/2607121915](plan/2607121915_arch-fix-linkgraph-wikilink-purity.md).
- **tax / nice-to-have** — a long tail of missing-dedicated-test debt on private helpers (covered transitively via integration-style tests), a handful of real duplication findings (e.g. `crossfilereferenceintegrity.toStringSlice` duplicating `settings.ToStringSlice`), and several oversized files that are split candidates. Left in the audit log per the audit workflow's own severity rubric — see the log for details.

Three resolved 2026-06 entries were archived to `docs/development/architecture-audit-archive.md` to keep the log under its file-length budget.

## Test plan

[...]

PR #737, opened 2026-07-12
perf: fix top 5 high-performance-go.md violations (round 6)

jeduden:
## Summary

An audit against `docs/development/high-performance-go.md` (dispatched as three parallel research agents scanning for regexp/fmt/string, allocation/prealloc, and struct-layout/data-structure anti-patterns) surfaced several violations of the project's performance guidelines that survived earlier perf-cleanup rounds (#723, #725, #729, #731, #732). This PR fixes the top 5, ranked by hotness and impact, each with red/green TDD and a benchmark or deterministic test pinning the win — then went through three rounds of xhigh-severity code review, one of which caught a real, `-race`-confirmed data race and fixed it before merge.

Three of the five findings share a pattern worth calling out: the shared `internal/integration/alloc_budget_test.go` fixture carries plain prose with no emphasis markup, no configured proper-name vocabulary, and empty schema/dedup maps, so it reports 0 allocs/op for these rules on `BenchmarkPerRuleAllocBudget` even though the real per-Check cost is substantial once the rule is actually configured/exercised the way it is in practice. Each fix below adds a dedicated, rule-local alloc test with a fixture that does exercise the cost, closing that blind spot for its rule going forward.

1. **MDS047 `ambiguous-emphasis`** — `scanLine` tracked the in-progress delimiter run through a `*emphRun` reassigned across loop iterations, forcing a fresh heap allocation for every delimiter run on a line. On ordinary emphasis-heavy prose (`**bold**`, `*italic*`, `_underscore_`) this cost ~10 allocs/line, invisible to the shared fixture because it carries no emphasis at all. Track the run by value plus an `open` bool instead, gate `scanLine` behind a cheap byte-count pre-check, skip building the escape/adjacency tracking maps when they can't possibly produce a diagnostic, and stop pre-sizing `diags` for a violation rate real documents rarely hit. Lands at **9 allocs/op** (was 41), under CLAUDE.md's documented ≤10 ceiling.

2. **`internal/rules/linelength` `wrapTokens`** — the line-packing loop grew its output line with `cur += " " + u` on every packed unit, reallocating the backing array each step — the same anti-pattern the sibling function `buildWrapUnits` in the same file was already fixed for. Switched to `strings.Builder`. 11 → 6 allocs/op on a 10-unit line packed onto one output line.

3. **MDS050 `proper-names`** — `buildNameEntries()` rebuilt a fresh `[]byte` plus an ASCII-lowercase copy per configured name on every single `Check` call, even though `r.Names` is immutable for the life of the `Rule` instance — a per-*Rule* memoization case, not a per-file one. First cached behind a lazy atomic-pointer/mutex (mirroring MDS063's `cachedBannedSet`), then **redesigned after code review found that pattern unsafe**: `internal/rule.CloneInstance` clones a rule via a raw, unsynchronized field copy, which races a concurrent `Check` on the source instance populating that lazy cache — confirmed with `go test -race` every time, and a real, reachable scenario given how `pkg/mdsmith.Session` shares rule instances across concurrent `Check`/`CheckVersion` calls. The final design precomputes `entries` synchronously inside `ApplySettings`, before any `Check` or clone can touch the instance — plain, write-once fields, no synchronization needed, verified race-free. 43 → 2 allocs/op with 20 configured names.

4. **MDS020 `required-structure` + `internal/schema`** — three `map[K]bool` sets used purely for presence checks (every write `true`/`struct{}{}`, never explicitly cleared): `requiredstructure`'s `claimed` (schema-heading walk) and `visited` (`<?include?>` cycle detection), and `schema.dedupedCUEErrorDiags`'s `seen` (dedup key set). Added an `isClaimed` helper mirroring the one `internal/schema/validate.go` already has for the identical pattern.

5. **`internal/lint` `fenceInfo`** — `char byte` sitting before `indent int`/`length int` cost 7 bytes of padding, landing the struct at 32 bytes. `openingFence` returns it by value on the hottest per-line scan path (`tryFence`, tried first for every non-blank line of every file via `scanBlock`'s dispatch), so the padding is copied repeatedly. Reordered large-to-small to 24 bytes, pinned by a new `unsafe.Sizeof` test; also dropped a redundant extra copy in `advanceFenceState` found in review.

## Code review

Three rounds of xhigh-severity review (8 independent finder angles per round), each addressed before the next:

- **Round 1** found and fixed: the MDS050 data race described above; an MDS047 alloc-budget test that pinned 15 allocs/op without going through CLAUDE.md's documented grandfather mechanism (now 9, under the ≤10 ceiling); two codecov patch-coverage gaps (`nextUnclaimed`'s new branch, `dedupedCUEErrorDiags`'s dedup branch); and a redundant struct copy in `advanceFenceState`.
- **Round 2** re-verified every round-1 fix empirically (including re-running the race probe against the redesigned code) and found no new correctness or concurrency bugs; it flagged one process issue (a commit that bundled 4 unrelated fixes, since split into 4 separate commits) plus two informational, no-action notes.
- **Round 3** re-verified the entire diff from scratch, ran the full build/vet/test/race/lint/`mdsmith check` matrix directly rather than trusting commit messages, and found no correctness or concurrency issues — only minor documentation nits.

## Test plan

- [x] `go build ./...`
- [x] `go test -race` on every touched package, including an adversarial `CloneInstance`-vs-`Check` race probe for the MDS050 fix
- [x] `go vet ./...` on every touched package
- [x] `go run ./cmd/mdsmith check .` (548 files, 0 failures)
- [x] Codecov patch coverage: 100% of modified/coverable lines
- [x] `internal/integration` per-rule alloc-budget gate still green for every touched rule (MDS047, MDS020)
- [x] Three rounds of xhigh code review, all findings addressed

---
_Generated by [Claude Code](https://claude.ai/code/session_01NgVEsnKFFaoHjwN3JbvkNy)_

### DIFF
--- a/internal/lint/layer0_fence.go
+++ b/internal/lint/layer0_fence.go
@@ -82,8 +82,7 @@ func closingFence(line []byte, fi fenceInfo) bool {
 func advanceFenceState(open *fenceInfo, line []byte, fi fenceInfo, opensFence bool) *fenceInfo {
 	if open == nil {
 		if opensFence {
-			f := fi
-			return &f
+			return &fi
 		}
 		return nil
 	}
--- a/internal/lint/layer0_fence.go
+++ b/internal/lint/layer0_fence.go
@@ -2,11 +2,15 @@ package lint
 
 import "bytes"
 
-// fenceInfo describes an opening fenced-code fence line.
+// fenceInfo describes an opening fenced-code fence line. Fields are
+// ordered large-to-small (both ints before the two single-byte
+// fields) to avoid the padding a byte-then-int layout would cost;
+// openingFence returns this by value on the hottest per-line scan
+// path (docs/development/high-performance-go.md "Struct layout").
 type fenceInfo struct {
-	char   byte
 	indent int
 	length int
+	char   byte
 	// hasInfo records whether the opening fence carries a non-empty info
 	// string after the fence run. goldmark exposes no source position for
 	// an info-less, content-less fence, so the projection emits no lines
--- a/internal/lint/layer0_fence_test.go
+++ b/internal/lint/layer0_fence_test.go
@@ -1,0 +1,22 @@
+package lint
+
+import (
+	"testing"
+	"unsafe"
+)
+
+// TestFenceInfo_FieldOrderMinimizesPadding pins fenceInfo's memory
+// layout at 24 bytes. openingFence returns fenceInfo by value on the
+// hottest per-line scan path (tryFence, tried first for every
+// non-blank line of every file), so struct padding is copied
+// repeatedly. Ordering fields large-to-small (both ints, then the two
+// single-byte fields) avoids the 8 bytes of padding a byte-then-int
+// layout costs. See docs/development/high-performance-go.md
+// "Struct layout" / "Order fields large-to-small".
+func TestFenceInfo_FieldOrderMinimizesPadding(t *testing.T) {
+	got := unsafe.Sizeof(fenceInfo{})
+	if got > 24 {
+		t.Fatalf("unsafe.Sizeof(fenceInfo{}) = %d, want <= 24 (large-to-small field order); "+
+			"see docs/development/high-performance-go.md", got)
+	}
+}
--- a/internal/rules/ambiguousemphasis/alloc_test.go
+++ b/internal/rules/ambiguousemphasis/alloc_test.go
@@ -1,0 +1,87 @@
+package ambiguousemphasis
+
+import (
+	"testing"
+
+	"github.com/jeduden/mdsmith/internal/lint"
+	"github.com/stretchr/testify/require"
+)
+
+// allocBudgetMDS047 pins Check's allocation cost on ordinary prose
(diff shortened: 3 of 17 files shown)


jeduden:
🟢 **Merge Queue** — picked up

**Next:** No action needed — you'll get another comment when CI starts on the batch. [View merge queue run](https://github.com/jeduden/mdsmith/actions/runs/29209710158).

jeduden:

Merged into batch branch [`merge-queue/batch-736-1783891778`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-736-1783891778) alongside #736. [View CI run](https://github.com/jeduden/mdsmith/actions/runs/29209750288).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`fd27822`](https://github.com/jeduden/mdsmith/commit/fd278220bc98337b11c33a4c480803804c86321d). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/29209750288).

PR #735, opened 2026-07-12
review: fix double-scan, allocation regression, and test gaps

jeduden:
## Summary

Follow-up fixes from the round-2 `xhigh` code review of the astutil whitespace-helpers consolidation (PR #734).

- **orderedlistnumbering/checkListLines, checkItem**: revert `strconv.Itoa` + string concatenation back to `fmt.Sprintf`. The concat form makes 3–4 allocations per diagnostic versus Sprintf's ~1 via its pooled `pp` — a confirmed allocation regression.

- **astutil_test/TestCollectSectionHeadings_Memoized**: add `require.Len` guard on the second `CollectSectionHeadings` result so a broken memoization path fails with a clear assertion message rather than a panic on `h2[0]`.

- **astutil_test/TestCountLeadingSpaces**: add the space-then-tab boundary case (`" \tabc"` → `1`) to the canonical unit test. This case existed in `listindent/helpers_test.go` but not in `astutil`'s own test, leaving a gap where a cutset change from `" "` to `" \t"` would pass astutil's tests undetected.

PR #739, opened 2026-07-13
Fix linkgraph purity-contract mismatch for wikilink resolution

jeduden:
## Summary

Resolves plan `2607121915`: `go.md` documented `internal/linkgraph` as "pure (no file reads, no workspace walks)" but `NewWikilinkIndex` and `ResolveWikiLink` both called `fs.WalkDir`, and they maintained two independent copies of the walk/match algorithm.

- **Deduplication**: `ResolveWikiLink` now delegates to `NewWikilinkIndex(root).Resolve(target)` — the walk/match algorithm lives in one place.
- **Doc accuracy**: `go.md`'s `internal/linkgraph` entry now correctly scopes the purity guarantee to per-file extraction functions and documents `NewWikilinkIndex`, `ResolveWikiLink`, and `WikilinkIndexFor` as workspace-walking (with `WikilinkIndexFor` memoizing per run via `lint.RunCache`).
- **Stale comments fixed**: `NewWikilinkIndex`'s comment no longer claims a `ResolveWikiLink` fallback (circular after the refactor); `wikilinkSearchKey`'s comment now references `WikilinkIndex.Resolve` and documents its pre-normalized-input contract.
- **Dead code removed**: `_ = from` in `ResolveWikiLink` (Go function parameters don't need silencing); redundant `strings.ReplaceAll` backslash normalization in `wikilinkSearchKey` (already done by `Resolve` before calling it).
- **Test updated**: `TestWikilinkSearchKey` drops the raw-backslash case (now belongs at the `Resolve` level, which already has it in `TestWikilinkIndex_ResolveSemantics`).

## Test plan

- [ ] `go build ./...` passes
- [ ] `go test ./internal/linkgraph/... ./cmd/mdsmith/... ./internal/rules/crossfilereferenceintegrity/...` passes
- [ ] `go test ./...` passes
- [ ] `mdsmith check .` — 0 failures

---
_Generated by [Claude Code](https://claude.ai/code/session_018boGj6rmVRTQPFqEYKK1rb)_

PR #738, opened 2026-07-12
test: add dedicated unit tests for unexported helpers (plan 2607051919)

jeduden:
## Summary

Implements plan `2607051919_arch-fix-new-helper-tests.md`: gives every unexported helper flagged by the 2026-07-05 architecture audit its own named `TestFoo` test. Previously these helpers were covered only transitively through end-to-end caller tests.

## Changes

**`internal/wordlist/wordlist_test.go`** — `TestAllNames` (empty, single, sorted) and `TestFlatten` (single-entry, extends-chain, cycle, unknown-parent).

**`internal/config/wordlist_files_test.go`** (created) — `TestToWordlistMap`, `TestStripLists`, `TestResolveListEntries`, `TestExpandRuleLists`; these four helpers had zero test-file references before this PR.

**`internal/convention/nollmtells_test.go`** (created) — `TestLlmVocabulary`, `TestLlmPhrases`, `TestLlmVocabularyAndPhrases`, `TestLlmParagraphOpeners`; each asserts non-empty output and a couple of known entries, mirroring `TestNoLLMTellsWordlists`.

**`internal/rules/linelength/fix_test.go`** — `TestOverlapsGeneratedRange`, `TestTrimTrailingCR`, `TestCloneBytes`.

**`internal/rules/linelength/reflow_test.go`** — `TestParagraphHasRawHTML` (without HTML and with inline raw HTML node).

**`internal/rules/concisenessscoring/classifier/model_test.go`** — `TestBuildPhraseMarkers` (empty input, applies phrase marker).

**`plan/2607051919_arch-fix-new-helper-tests.md`** — status marked ✅, all acceptance criteria checked.

## xhigh Code-Review Fixes (3 rounds, opus)

- `nollmtells_test.go`: `assert.Len` → `require.Len` before `got[0]` access so a length mismatch produces a clean test failure instead of a panic.
- `reflow_test.go`: removed dead `doc` parent-link scaffolding from both `TestParagraphHasRawHTML` tests (`paragraphHasRawHTML` never calls `Parent()`).
- `model_test.go`: deleted `TestBuildPhraseMarkers_LengthPreserving`; its `len(got)==2` assertion was already fully covered by `TestBuildPhraseMarkers_AppliesPhraseMarker`.

## Test plan

- `go test ./...` — green
- `go run ./cmd/mdsmith check .` — 0 failures

---
_Generated by [Claude Code](https://claude.ai/code/session_015qYnUAc5XUmrQBwiEYyd4j)_

PR #740, opened 2026-07-13
perf: fix top 5 high-performance-go.md violations (round 8)

jeduden:
## Summary

An audit against `docs/development/high-performance-go.md` (dispatched as four parallel research agents scanning for regexp/allocation, memoization/skip-work, struct-layout/data-structure, and concurrency/misc anti-patterns) surfaced several violations that survived seven earlier perf-cleanup rounds (#723, #725, #729, #731, #732, #735, #737). Prior rounds had already cleared `golangci-lint`'s `perfsprint`/`prealloc`/`gocritic`(performance) checks to 0 issues repo-wide, so this round's findings are all outside what those linters catch. This PR fixes the top 5, ranked by hotness and impact, each with red/green TDD — then went through three rounds of xhigh-severity code review, each of which found and fixed real issues before the next round.

1. **`internal/lint.File` struct size (688 → 640 bytes), plus a residual ptrdata refinement** — `File` is allocated once per `NewFile` call, the single most common allocation across the engine's per-file `Check` path. Its 24 lazy-init guard fields (10 `atomic.Bool` + 10 `sync.Mutex` + one `sync.Once`) were each declared beside the cache field they protect, interleaving 4-byte-aligned scalars between 8-byte-aligned pointer/slice/map fields and forcing a padding gap after nearly every guard pair. Grouped every guard into one block after all pointer-bearing fields — each guard's comment still names the cache field it protects, only the declaration site moved. Guarded by both a pinned `unsafe.Sizeof` budget and a `structlayout.AssertPointerFieldsFirst` ordering assertion (round-2 review added the latter; the size test alone couldn't catch a future field-order regression). A further round-2 finding shaved a residual 16 bytes of GC ptrdata: ending the pointer-bearing field block on a slice (`linkRefs`) instead of a bare pointer (`RunCache`) lets the slice's free len/cap tail fall outside the GC scan region.

2. **`internal/rules/astutil.SectionParagraph` ptrdata (32 → 24 bytes)** — `CollectSectionParagraphs` allocates one `SectionParagraph` per document paragraph, backing MDS023 (paragraph-readability) and MDS024 (paragraph-structure), both default-enabled — likely the highest-frequency struct construction in the rule set. `Line` (a plain int) was declared before `Node`/`Text`, its only pointer-bearing fields, so the GC's per-word ptrdata scan covered `Line`'s word too. Moved `Node`/`Text` first.

3. **`internal/rules/tablereadability` `table`/`tableRow` ptrdata (16 → 8 bytes each)** — the sibling `tablefmt` package already reordered its near-identical `table`/`row` structs for this exact reason (commit `9284744f`), but `tablereadability`'s local copy — kept separate for its zero-alloc `[][]byte` cell representation — was never updated to match. `parseTables` allocates one of each per table/row MDS026 scans.

4. **`internal/schema.ValidateIndex` read path** — read the on-disk `<?catalog?>`/index side-output via a raw `os.ReadFile`, the only cross-file read in the package that bypassed the `bytelimit` convention every sibling call site follows. Routed through `bytelimit.ReadFileLimited`, passing `f.MaxInputBytes` straight through — matching that field's documented "zero or negative means unlimited" contract and every other consumer of it. (Round-1 review caught an intermediate version of this fix that substituted a 2 MB default whenever `MaxInputBytes` was `<= 0`, silently breaking that contract for any caller — LSP server, direct `pkg/mdsmith.Session`, `--max-input-size 0` — that asked for no cap; fixed before it shipped.)

5. **`pkg/markdown/flavor/ext` abbreviation matching** — `findMatches` probes every word-boundary candidate position in a Text node's body against every defined abbreviation term, and `bestMatchAt` converted each term from its map key to `[]byte` on *every single* (position, term) probe — the dominant allocator on any document with a defined abbreviation. Precomputed each term's `[]byte` form once per document: `tableTerms(table)` is called once in `Transform`, and the resulting slice threads through `rewriteText`/`findMatches`/`bestMatchAt` for every Text node in that parse. (Round-1 review caught an intermediate version that recomputed this once per `findMatches` call — i.e. once per Text node rather than once per document — before it shipped; round-2 review added a dedicated `TestFindMatches_AllocBudget` to guard the corrected scope going forward.)

## Code review

Three rounds of xhigh-severity review (8 independent finder angles per round), each addressed before the next:

- **Round 1** found and fixed: the `MaxInputBytes`/`ValidateIndex` contract bug described above; an unreachable defensive branch in `findMatches` added without a driving test (also the direct cause of a codecov/patch failure); the `tableTerms` per-Text-node-instead-of-per-document scoping gap; and a missing `structlayout.AssertPointerFieldsFirst` test for `File` (added alongside the existing size budget).
- **Round 2** re-verified every round-1 fix from scratch and found: a backwards doc comment in `file_size_test.go` (fixed), a residual 16-byte GC-ptrdata optimization in `File` (fixed, described above), and evaluated two "could go further" suggestions it explicitly framed as optional follow-up rather than blocking — a package-wide inconsistency between two *different*, differently-documented fields' zero-value policies (`FileReader.MaxBytes` vs `lint.File.MaxInputBytes`), and storing abbreviation `[]byte` term forms at definition time instead of `Transform` time — both deferred as out of this round's scope. It also flagged a test fixture that was intentionally kept oversized after evaluation: shrinking it would have let it pass under both the correct and the previously-buggy behavior, defeating its purpose as a regression guard.
- **Round 3** re-verified the full build/vet/test/lint/`mdsmith check`/`fieldalignment` matrix from scratch across all 10 commits and found no further correctness issues; it did flag that this PR description itself had drifted out of sync with the final code after two rounds of fix-up commits, which is what this text now corrects.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go run ./cmd/mdsmith check .` (549 files, 0 failures)
- [x] `internal/integration` per-rule alloc-budget gate still green for every rule
- [x] `fieldalignment` re-run after the final commit: no remaining findings for `File`, `SectionParagraph`, `table`, or `tableRow`
- [x] Codecov patch coverage: 100% of modified/coverable lines
- [x] Each fix's red test confirmed failing before the corresponding implementation change, and passing after
- [x] Three rounds of xhigh code review, all findings addressed or explicitly deferred with rationale

---
_Generated by [Claude Code](https://claude.ai/code/session_016cg8TcTdbY2cVaHaMUKyjX)_

### DIFF
--- a/internal/lint/file.go
+++ b/internal/lint/file.go
@@ -13,7 +13,18 @@ import (
 	"github.com/jeduden/mdsmith/pkg/markdown"
 )
 
-// File holds a parsed Markdown document and its source.
+// File holds a parsed Markdown document and its source. File is
+// allocated once per NewFile call — the single most common
+// allocation across the engine's per-file Check path — so its field
+// order follows docs/development/high-performance-go.md#struct-layout
+// large-to-small: every pointer/slice/map/interface field first,
+// then the block of 4-byte-aligned lazy-init guards (gitignoreOnce,
+// every *Done atomic.Bool, every *Mu sync.Mutex, and the two bools),
+// then the two remaining 8-byte scalars last. Each guard field still
+// documents which cache slice/map above it protects; only the
+// declaration site moved, to let same-size fields pack without the
+// per-pair padding that interleaving them costs. See
+// file_size_test.go for the pinned budget.
 type File struct {
 	Path        string
 	Source      []byte
@@ -23,26 +34,6 @@ type File struct {
 	RootFS      fs.FS
 	RootDir     string
 	FrontMatter []byte
-	LineOffset  int
-
-	// StripFrontMatter records whether this file was parsed in
-	// front-matter-stripping mode. Rules that read other files
-	// from the corpus should mirror the same mode so that line
-	// numbers in cross-file diagnostics are computed against the
-	// same coordinate system as the current file.
-	StripFrontMatter bool
-
-	// DryRun, when true, signals that the surrounding fix run must
-	// not touch the filesystem or the git index. Fixable rules whose
-	// Fix method has side effects beyond returning the new file
-	// bytes (e.g. writing a sibling repo file, staging via git)
-	// must check this flag and skip the side effect.
-	DryRun bool
-
-	// MaxInputBytes is the maximum file size in bytes that rules
-	// should enforce when reading secondary files (includes, schemas,
-	// cross-references). Zero or negative means unlimited.
-	MaxInputBytes int64
 
 	// GitignoreFunc is a lazy factory for the gitignore matcher.
 	// It is called at most once (on first access via GetGitignore)
@@ -50,7 +41,6 @@ type File struct {
 	// never trigger matcher construction. sync.Once keeps the lazy
 	// build race-free if a *File is shared across goroutines.
 	GitignoreFunc func() *gitignore.Matcher
-	gitignoreOnce sync.Once
 	gitignoreVal  *gitignore.Matcher
 
 	// GeneratedRanges records the content line ranges of generated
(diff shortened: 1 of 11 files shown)


jeduden:
🟢 **Merge Queue** — picked up

**Next:** No action needed — you'll get another comment when CI starts on the batch. [View merge queue run](https://github.com/jeduden/mdsmith/actions/runs/29557190009).

jeduden:

Merged into batch branch [`merge-queue/batch-738-1784266050`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-738-1784266050) alongside #738, #739. [View CI run](https://github.com/jeduden/mdsmith/actions/runs/29557408649).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`5124cbc`](https://github.com/jeduden/mdsmith/commit/5124cbc695b13828909b145375e372ea2b52ff79). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/29557408649).

PR #742, opened 2026-07-16
perf: fix top 5 high-performance-go.md violations (round 7)

jeduden:
## Summary

An audit against `docs/development/high-performance-go.md` (dispatched as three parallel research agents scanning for allocation/string, data-structure/slice, and concurrency/misc anti-patterns) confirmed the codebase is largely clean after six prior perf-cleanup rounds (#723, #725, #729, #731, #732, #737). This round found only five candidates total across all three scans — every finding below is genuine and empirically proven with a git-stash A/B `testing.AllocsPerRun` comparison, each backed by red/green TDD (a test that fails against the unfixed code and passes against the fix).

One candidate the initial scan flagged (`tableformat`'s `normalizeEdges` doing a `string(line)` conversion) turned out to be a **false positive**: A/B testing showed zero allocation difference, because Go's cross-function escape analysis already eliminates the cost when the callee never retains the argument. It was dropped rather than "fixed" without a measurable win, and a genuine replacement was found and verified the same way.

1. **MDS030 `empty-section-body`** — `topLevelNodes` pre-sized its per-file `[]ast.Node` slice with a fixed heuristic capacity of 8. `root.ChildCount()` is an O(1) field goldmark already tracks per node, so sizing from it allocates exactly once regardless of document length instead of paying `append`'s grow+copy on documents with more top-level nodes. MDS030 is enabled by default, so this runs on every file's `Check`. Verified 3 → 1 allocs/op on a 30-node fixture.

2. **`required-structure`'s `resolveBodySyncLine`** (the `Fix`-path counterpart to `checkBodySync`) — scanned every candidate line with `strings.TrimSpace(string(work[i]))`, even on the common non-matching case. Switched to `bytes.TrimSpace` / `bytes.Equal` / `regexp.Match` against a precomputed `expectedBytes`, the same convention `checkBodySync` already uses. Verified 27 → 7 allocs/op on a 20-line non-matching fixture.

3. **`single-h1`'s `Fix`** — built the `reps` replacement-list slice by appending from `nil`, even though `len(toDemote)` is a hard upper bound known before the loop starts (each iteration appends at most one entry). Pre-sizing avoids the `append` grow+copy. Verified 48 → 44 allocs/op on a 19-extra-H1 fixture.

4. **`internal/rulelayer`** — `astProjectionConsumers` and `knownNilASTSafe` are presence-only sets (every read is a membership check; no code path reads a stored `false` as distinct from absence), declared as `map[string]bool`. Converted to `map[string]struct{}`, the documented zero-byte-value idiom for a set. This is a memory-footprint fix rather than an allocation-count win (both maps hold 0–1 entries in production), backed by the existing correctness test suite rather than a new alloc test.

5. **MDS066 `commands-show-output`'s `stripPromptAfter`** — copied every content line through `string(line)` on all three "nothing to strip" branches, even though the value is only ever written back verbatim by the caller. Blank separator lines between commands inside a flagged `$`-prompt block hit this branch routinely. Returning the `[]byte` line directly on the unchanged branches costs nothing; the strip branch now builds its result with `make`+`append` instead of string concatenation, matching the new `[]byte` return type. Verified 1 → 0 allocs/op via a dedicated unchanged-line fixture.

## Test plan

- [x] `go build ./...`
- [x] `go test ./...` (full suite green, 158/158 packages)
- [x] `go test -race` on the touched `commandsshowoutput` package
- [x] `go vet ./...` on every touched package
- [x] `go tool -modfile=tools/go.mod golangci-lint run` on every touched package (0 issues)
- [x] `go run ./cmd/mdsmith check .` (549 files, 0 failures)
- [x] Every fix's alloc test verified red (fails on unfixed code via `git stash`) before being verified green

---
_Generated by [Claude Code](https://claude.ai/code/session_01NGJRXrA9e4QzZvYvgCqSws)_

### DIFF
--- a/internal/rulelayer/rulelayer.go
+++ b/internal/rulelayer/rulelayer.go
@@ -72,8 +72,9 @@ var layerByID = buildLayerMap()
 //
 // The map stays as the seam for any future projection-only consumer the
 // audit marks A-no-skipping but Layer N does not yet back. It is empty
-// today.
-var astProjectionConsumers = map[string]bool{}
+// today. map[string]struct{} because membership is all that is ever
+// checked — no entry's value is read.
+var astProjectionConsumers = map[string]struct{}{}
 
 // knownNilASTSafe lists rules confirmed nil-AST-safe by code inspection but
 // whose walk-audit probe cannot fire them, so the audit classifies them
@@ -83,8 +84,10 @@ var astProjectionConsumers = map[string]bool{}
 // signals prove nothing. Adding a rule here is a manual commitment that its
 // Check never dereferences f.AST — keep it limited to rules whose manifest
 // entry has reads_file_ast: false (enforced by rulelayer_test.go).
-var knownNilASTSafe = map[string]bool{
-	"MDS069": true, // unique-frontmatter: reads f.FrontMatter + RunCache, never f.AST
+// map[string]struct{} because membership is all that is ever checked —
+// no entry's value is read.
+var knownNilASTSafe = map[string]struct{}{
+	"MDS069": {}, // unique-frontmatter: reads f.FrontMatter + RunCache, never f.AST
 }
 
 // buildLayerMap decodes the embedded manifest into the id→layer table.
@@ -110,7 +113,8 @@ func buildLayerMapFrom(manifest []byte) map[string]Layer {
 	}
 	m := make(map[string]Layer, len(entries))
 	for _, e := range entries {
-		if !astProjectionConsumers[e.ID] && nilASTBackable(e) {
+		_, isProjectionConsumer := astProjectionConsumers[e.ID]
+		if !isProjectionConsumer && nilASTBackable(e) {
 			m[e.ID] = Layer0
 		} else {
 			m[e.ID] = LayerAST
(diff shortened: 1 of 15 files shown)


jeduden:
🟢 **Merge Queue** — picked up

**Next:** No action needed — you'll get another comment when CI starts on the batch. [View merge queue run](https://github.com/jeduden/mdsmith/actions/runs/29557044768).

jeduden:

Merged into batch branch [`merge-queue/batch-741-1784265638`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-741-1784265638) alongside #741. [View CI run](https://github.com/jeduden/mdsmith/actions/runs/29557095870).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`e9ea9b0`](https://github.com/jeduden/mdsmith/commit/e9ea9b0fbc23c82a49583d23d26327b4e4035786). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/29557095870).

PR #748, opened 2026-07-19
docs(wordlists): complete plan 2606251522 — integration test and doc updates

jeduden:
## Summary

- Adds integration test `TestWordlistFiles_ExtendsChainResolvesAndDiagnosticFires` that sets up a workspace with two word-list files (`base.yaml` and `team.yaml` that extends base), configures `forbidden-text` to reference the `team` list, and asserts that MDS056 fires on "synergy" — a word inherited through the extends chain. This covers acceptance criterion #1 from plan 2606251522.
- Documents the `lists:` setting for placeholder rules in `docs/background/concepts/placeholder-grammar.md` — each opt-in rule implements `WordlistConsumer` with a `placeholders` target, allowing projects to share token vocabularies via named word-lists.
- Documents `lists:` support in convention files in `docs/reference/convention-files.md` — a convention can preset which word-lists a rule draws from.
- Adds a note to `docs/reference/conventions.md` that `lists:` always appends across config layers, with a link to the word-list files reference.
- Marks plan 2606251522 complete in both `plan/2606251522_user-extensible-wordlists.md` and `PLAN.md`.

## Test plan

- [ ] `go test ./internal/integration/...` — new `TestWordlistFiles_ExtendsChainResolvesAndDiagnosticFires` passes
- [ ] `go test ./...` — all tests pass
- [ ] `go run ./cmd/mdsmith check .` — all 556 docs pass (no MDS022/MDS023 regressions in edited files)

---
_Generated by [Claude Code](https://claude.ai/code/session_01Qk4dXhBDW1BF3ecCRA64gP)_

PR #752, opened 2026-07-19
fix(merge-driver): scope ignore-derived -merge lines to Markdown (#750)

### LINKED ISSUE #750: merge-driver install: ignore-derived '-merge' lines disable git 3-way merge for non-markdown files
## Description

`mdsmith merge-driver install` derives its `.gitattributes` exclude lines from the `ignore:` patterns in `.mdsmith.yml`, emitting one `<glob> -merge` line per pattern.

But `ignore:` is a **markdown-lint** scoping list — mdsmith's `files:` is `**/*.md` / `**/*.markdown`, so an ignore pattern can only ever be *intended* to affect markdown. The emitted `-merge` attribute, by contrast, is **not** scoped to markdown: it applies to every file matching the glob.

So a coarse directory ignore like `demos/**` — perfectly reasonable for grandfathering un-linted markdown — silently turns off git's 3-way merge for **all the source code in that tree**. Git then treats those files as binary for merging, and any two branches touching the same file conflict irreconcilably, even when their edits are nowhere near each other.

This is easy to miss because nothing in the lint output or the merge-driver output mentions non-markdown files, and `.gitattributes` is not a file most contributors read.

## Seen behavior

`.mdsmith.yml` grandfathers some directories from markdown lint:

```yaml
files:
  - '**/*.md'
  - '**/*.markdown'
ignore:
  - "demos/**"
  - "poc/**"
  - "shared/**"
```

`mdsmith merge-driver install` (v0.46.0) then writes:

```gitattributes
# BEGIN mdsmith merge-driver
*.md merge=mdsmith
*.markdown merge=mdsmith
...
demos/** -merge
poc/** -merge
shared/** -merge
...
# END mdsmith merge-driver
```

and git reports every file in those trees — not just markdown — as unmergeable:

```console
$ git check-attr merge demos/world-engine/vrt.zig
demos/world-engine/vrt.zig: merge: unset
```

Minimal repro of the consequence (two branches editing *disjoint* regions of one 18k-line Zig file — line 3 and EOF):

```console
$ git checkout -b laneA && sed -i '3i // LANE A' demos/world-engine/vrt.zig && git commit -qam laneA
$ git checkout -b laneB <base> && printf '\n// LANE B\n' >> demos/world-engine/vrt.zig && git commit -qam laneB
$ git merge laneA
Auto-merging demos/world-engine/vrt.zig
CONFLICT (content): Merge conflict in demos/world-engine/vrt.zig
Automatic merge failed; fix conflicts and then commit the result.
```

One of the two edits is lost unless it is merged by hand. With the `-merge` line removed, the identical merge succeeds and both edits survive.

In our repo the affected file appears in 83 of the last 200 commits, so this fires constantly once more than one branch is in flight.

## Expected behavior

The merge driver should never change git's merge behavior for files it would never handle. The derived exclude lines should be **intersected with the driver's include set**, i.e. for `ignore: ["demos/**"]` with the default include globs, emit:

```gitattributes
demos/**/*.md -merge
demos/**/*.markdown -merge
```

Turning off the mdsmith driver for markdown in an ignored tree is the intent; turning off git's default text merge for that tree's source code is a side effect nobody asked for.

Failing that, either of these would also resolve it:

- emit `merge=text` rather than `-merge` for the exclusions, so the driver is disabled but git's normal 3-way merge is preserved; or
- keep the current behavior but make it opt-in, and warn at `install` time when a derived exclude glob matches files outside the include set.

## Workaround

Appending a re-enable block *after* the managed block (last-match-wins, so it survives regeneration by `install` and by the MDS048 `git-hook-sync` auto-fix):

```gitattributes
# END mdsmith merge-driver

demos/**/*.zig merge
demos/**/*.glsl merge
```

The seemingly obvious config-side fix — scoping the ignore patterns themselves to `demos/**/*.md` — does **not** work, because `ignore:` doubles as the lint scope: narrowing it un-grandfathers files and reintroduces lint failures (18 files, in our case). The two consumers of `ignore:` want different scopes, which is really the root of this.

Custom include globs passed to `merge-driver install` are also not a way out, since the docs note they are incompatible with the `git-hook-sync` auto-fix.

## Environment

- mdsmith v0.46.0 (pinned via mise, `"github:jeduden/mdsmith" = "0.46.0"`)
- git 2.x, Linux

jeduden:
## Summary

Fixes #750.

`mdsmith merge-driver install` derived its `.gitattributes` exclude lines from the `ignore:` patterns in `.mdsmith.yml`, emitting one bare ` -merge` line per pattern. But `ignore:` is a **markdown-lint** scoping list (`files:` is `**/*.md` / `**/*.markdown`), while a bare `demo/** -merge` attribute is **not** extension-scoped — it disabled git's 3-way merge for *every* file in that tree, including source code that nobody asked to grandfather. Two branches editing disjoint regions of the same non-Markdown file then conflicted irreconcilably (git treats the file as binary for merging).

This scopes each ignore-derived exclude to Markdown, then centralizes the Markdown-extension set so the fix (and every other extension check) has a single source of truth.

> **Note on the committed `.gitattributes`:** it deliberately stays in the **legacy bare form** in this PR. The merge queue validates `.gitattributes` with `merge-driver ci-install` run by the **pinned release baseline** (`setup-mdsmith-pinned-version`), which predates markdown scoping — so changing the committed file's format would fail the queue. `install` writes the new scoped form for users; the repository migrates its own `.gitattributes` in a follow-up once this ships in a release and the pin is bumped (see `docs/development/adopt-new-directive-syntax.md`).

### 1. `fix(merge-driver)`: scope ignore-derived `-merge` lines to Markdown (#750)

`githooks.GlobsFromConfig` now intersects each representable ignore pattern with the Markdown include extensions. For `ignore: ["demo/**"]` it emits:

```gitattributes
demo/**/*.md -merge
demo/**/*.markdown -merge
```

instead of a bare `demo/** -merge`. Every emitted exclude ends in a Markdown extension, so a derived `-merge` line can **never** match a non-Markdown file. A pattern that already targets a specific Markdown extension (e.g. `internal/rules/*/bad.md`) is kept verbatim. This is what `merge-driver install` writes.

Pattern handling:

| `ignore:` pattern | emitted excludes |
| --- | --- |
| `demo/**` | `demo/**/*.md`, `demo/**/*.markdown` |
| `vendor/*` | `vendor/*.md`, `vendor/*.markdown` |
| `internal/rules/*/bad.md` | `internal/rules/*/bad.md` (already Markdown — verbatim) |
| `build/` | `build/**/*.md`, `build/**/*.markdown` |

Verified with `git check-attr`: a non-Markdown file in an ignored tree (`demo/world.zig`) keeps git's default merge (`unspecified`), while Markdown in that tree (`demo/x.md`, `demo/x.markdown`) is `unset`. Unrepresentable patterns (`!` negation, whitespace) are still skipped and reported exactly as before.

### 2. `fix(merge-driver)`: handle trailing-slash ignore patterns

Self-review follow-up: a gitignore-style trailing-slash ignore (`build/`, `node_modules/`) has an empty final path segment and produced a malformed `build//**/*.md` double slash. Folded into the recursive-tree branch so it emits `build/**/*.md`; added a dedicated test.

### 3. `refactor(mdpath)`: make the Markdown extension set a single source of truth

The `.md`/`.markdown` set was duplicated across ~8 packages (discovery globs, the CLI walk, the merge-driver include set and discovery walk, the LSP document selector and file watcher, wikilink resolution, the link-style rule). Centralized it in `internal/mdpath`:

- `Extensions()` → `[".md", ".markdown"]`
- `FileGlobs()` → `["*.md", "*.markdown"]`
- `RecursiveGlobs()` → `["**/*.md", "**/*.markdown"]`
- `HasMarkdownExt()` → case-insensitive, zero-alloc membership test

`config.DefaultFiles`, `lint.isMarkdown`, `githooks.DefaultIncludes`/`GlobsFromConfig`/`DiscoverFiles`, `linkgraph` wikilink resolution, `linkstyle`, and the LSP `indexPatterns`/`isMarkdownExt`/file-watcher now all derive from these helpers. **No change to which extensions are recognized**; adding a new one (e.g. `.mdx`) is now a one-line change in `mdpath`.

### 4. `refactor(mdpath)`: use `slices.Clone` in `Extensions`

Match the repo's established clone pattern. No behavior change.

### 5. `fix(merge-driver)`: keep the committed `.gitattributes` on the legacy format for the pinned queue

The merge queue's `ci-install` runs the **pinned** v0.41.0 baseline, which expects the bare `demo/** -merge` render; regenerating the committed file to the scoped form made it report drift and fail the queue. Reverted the committed `.gitattributes` to the legacy bare form (so the pinned baseline stays green — for this PR and every later one, since main never changes format), added `githooks.LegacyGlobs` (the pinned-baseline render), and repointed the dogfood `TestRepoGitattributesInSyncWithConfig` at it. `GlobsFromConfig` still renders the scoped form that `install` writes.

## Testing

- New unit tests for the scoping helpers (`scopeExcludeToMarkdown`, `splitLastSegment`, trailing-slash, markdown-scoped invariant), the `mdpath` helpers (`Extensions`, `FileGlobs`, `RecursiveGlobs`, `HasMarkdownExt`), and `LegacyGlobs`.
- Updated the `GlobsFromConfig`/`LoadGlobs`, MDS048 rule, and CLI install tests to the scoped output; the dogfood test verifies the committed legacy `.gitattributes` against `LegacyGlobs`.
- `go test ./...`, `go vet ./...`, `golangci-lint` (0 issues), `gofmt`, the `mdpath` 0-alloc gate, and `mdsmith check .` (556 files, 0 failures) all pass. Codecov: all modified lines covered.

https://claude.ai/code/session_01NK7yWgvWUN5xD1NTuGVZvZ

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -4,14 +4,23 @@
 # BEGIN mdsmith merge-driver
 *.md merge=mdsmith
 *.markdown merge=mdsmith
-.git/** -merge
-demo/** -merge
-internal/rules/*/bad/** -merge
+.git/**/*.md -merge
+.git/**/*.markdown -merge
+demo/**/*.md -merge
+demo/**/*.markdown -merge
+internal/rules/*/bad/**/*.md -merge
+internal/rules/*/bad/**/*.markdown -merge
 internal/rules/*/bad.md -merge
-internal/rules/*/good/** -merge
-internal/rules/*/fixed/** -merge
-internal/rules/*/pattern/** -merge
-.claude/worktrees/** -merge
-editors/**/node_modules/** -merge
-editors/**/dist/** -merge
+internal/rules/*/good/**/*.md -merge
+internal/rules/*/good/**/*.markdown -merge
+internal/rules/*/fixed/**/*.md -merge
+internal/rules/*/fixed/**/*.markdown -merge
+internal/rules/*/pattern/**/*.md -merge
+internal/rules/*/pattern/**/*.markdown -merge
+.claude/worktrees/**/*.md -merge
+.claude/worktrees/**/*.markdown -merge
+editors/**/node_modules/**/*.md -merge
+editors/**/node_modules/**/*.markdown -merge
+editors/**/dist/**/*.md -merge
+editors/**/dist/**/*.markdown -merge
 # END mdsmith merge-driver
--- a/.gitattributes
+++ b/.gitattributes
@@ -4,23 +4,14 @@
 # BEGIN mdsmith merge-driver
 *.md merge=mdsmith
 *.markdown merge=mdsmith
-.git/**/*.md -merge
-.git/**/*.markdown -merge
-demo/**/*.md -merge
-demo/**/*.markdown -merge
-internal/rules/*/bad/**/*.md -merge
-internal/rules/*/bad/**/*.markdown -merge
+.git/** -merge
+demo/** -merge
+internal/rules/*/bad/** -merge
 internal/rules/*/bad.md -merge
-internal/rules/*/good/**/*.md -merge
-internal/rules/*/good/**/*.markdown -merge
-internal/rules/*/fixed/**/*.md -merge
-internal/rules/*/fixed/**/*.markdown -merge
-internal/rules/*/pattern/**/*.md -merge
-internal/rules/*/pattern/**/*.markdown -merge
-.claude/worktrees/**/*.md -merge
-.claude/worktrees/**/*.markdown -merge
-editors/**/node_modules/**/*.md -merge
-editors/**/node_modules/**/*.markdown -merge
-editors/**/dist/**/*.md -merge
-editors/**/dist/**/*.markdown -merge
+internal/rules/*/good/** -merge
+internal/rules/*/fixed/** -merge
+internal/rules/*/pattern/** -merge
+.claude/worktrees/** -merge
+editors/**/node_modules/** -merge
+editors/**/dist/** -merge
 # END mdsmith merge-driver
--- a/cmd/mdsmith/mergedriver.go
+++ b/cmd/mdsmith/mergedriver.go
@@ -32,8 +32,13 @@ Subcommands:
         Register the merge driver in git config and ensure
         .gitattributes assigns it. The managed block uses globs
         derived from the project's .mdsmith.yml: include patterns
-        (default: *.md and *.markdown) followed by an exclude line
-        for each ignore pattern (last-match-wins overrides).
+        (default: *.md and *.markdown) followed by markdown-scoped
+        exclude lines for each ignore pattern (last-match-wins
+        overrides). Each ignore pattern is intersected with the
+        markdown include extensions, so an ignore of demo/** emits
+        demo/**/*.md -merge and demo/**/*.markdown -merge rather
+        than a bare demo/** -merge that would also disable git's
+        merge for non-markdown files in that tree.
 
         Optional positional args replace the default include set
         when callers want to scope the merge driver to a custom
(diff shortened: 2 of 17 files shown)


jeduden:
🟢 **Merge Queue** — picked up

**Next:** No action needed — you'll get another comment when CI starts on the batch. [View merge queue run](https://github.com/jeduden/mdsmith/actions/runs/29704129960).

jeduden:

Merged into batch branch [`merge-queue/batch-748-1784495975`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-748-1784495975) alongside #748. [View CI run](https://github.com/jeduden/mdsmith/actions/runs/29704170066).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`431316a`](https://github.com/jeduden/mdsmith/commit/431316a7b794cf1570853a5cb34efe2c71e2a9e9). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/29704170066).

PR #760, opened 2026-07-20
fix(cli): skip non-Markdown files named explicitly on check/fix

### LINKED ISSUE #759: check/fix lint non-Markdown files given as explicit paths, and fix edits .gitattributes
## Summary

When a **non-Markdown file is passed explicitly** on the command line, `mdsmith check` lints it as Markdown and `mdsmith fix` **modifies it**. The directory walk gets this right and skips such files, so the bug only shows on an explicit path.

It bites hardest on `.gitattributes`, because mdsmith *generates* the merge-driver block in that file — so the tool ends up rewriting a file it owns, into a state that still fails its own check.

## Reproduce

In a repo with the mdsmith merge-driver block installed (mdsmith **v0.46.0**):

```console
$ mdsmith check .                 # directory walk — correct
stats: checked=143 fixed=0 failures=0 unfixed=0

$ mdsmith check .gitattributes    # explicit path — lints it as Markdown
stats: checked=1 fixed=0 failures=14 unfixed=14
```

The 14 diagnostics are Markdown **content** rules applied to a git config file:

```
      2 MDS001   (line too long)
      9 MDS013
      2 MDS017   (heading should not end with punctuation ".")
      1 MDS023   (paragraph too hard to read; readability index 23.1, max 14.0)
```

`#` comments are being parsed as ATX **headings**, and comment prose is being scored for **readability**:

```
.gitattributes:21:1  MDS017  heading should not end with punctuation "."
.gitattributes:3:1   MDS023  paragraph too hard to read (readability index: 23.1,
                             max 14.0); avg sentence length 30 words
```

## The damaging part: `fix` edits the file

```console
$ mdsmith fix .gitattributes
stats: checked=1 fixed=1 failures=14 unfixed=5
```

It applies *blank-line-around-headings* to a git config file, inserting blank lines between comment lines — including **inside mdsmith's own generated block**:

```diff
 # BEGIN mdsmith merge-driver
+
 *.md merge=mdsmith
 *.markdown merge=mdsmith
@@
 zig-pkg/** -merge
+
 # END mdsmith merge-driver

 # Re-enable git 3-way merge for CODE under the trees mdsmith marks -merge.
+
 # mdsmith derives those -merge lines from .mdsmith.yml ignore patterns, ...
+
 # a MARKDOWN-lint grandfathering list; unscoped, they also disable merging ...
```

Note it is **not idempotent-to-clean**: after `fix`, 5 diagnostics remain unfixed, so the file is modified *and* still failing.

The blank lines are semantically harmless to git, but the practical cost is real: any `mdsmith fix <path>` that happens to name this file leaves an unexplained diff in the working tree. In our case it produced a spurious modification to `.gitattributes` during unrelated work, which is easy to commit by reflex.

## Expected

`check`/`fix` should skip files that are not Markdown, regardless of whether the path was discovered by the walk or named explicitly — matching the walk's existing behaviour. (`mdsmith kinds resolve .gitattributes` already reports `effective kinds: (none)`.)

At minimum, mdsmith should never apply Markdown content rules to the `.gitattributes` block it generates itself.

## Environment

- mdsmith **v0.46.0** (pinned via mise: `"github:jeduden/mdsmith" = "0.46.0"`)
- Repo: `jeduden/smalt`

## Secondary observation (not currently reproducible)

Separately, I saw `MDS048 .gitattributes managed block is out of sync` where the *`has`* form was markdown-scoped (`demos/**/*.md`, `demos/**/*.markdown`) and the *`should have`* form was unscoped (`demos/**`) — i.e. a `fix`-produced block that `check` then rejected. The block is in sync now, so I cannot reproduce it on demand and cannot say which side wrote the scoped form. Recording it only in case it is a known fix/check disagreement; the primary bug above stands on its own.

jeduden:
## Summary

Fixes #759. When a non-Markdown file was named **explicitly** on the command line, `mdsmith check` linted it as Markdown and `mdsmith fix` **rewrote** it. The directory walk and glob expansion already skip such files; only the explicit-path branch was missing the gate.

The damage was worst on `.gitattributes`, because mdsmith *generates* the merge-driver block there — so `fix` rewrote a file it owns (inserting blank lines around `#` comment lines it mis-parsed as headings), and the result was not even idempotent-to-clean.

## Root cause

`internal/lint/files.go` routes each positional argument three ways, but only two gated on `isMarkdown`:

- **directory walk** (`walkDir`) — gated ✅
- **glob expansion** (`resolveGlob`) — gated ✅
- **explicit file path** (`resolveArg`) — added the file **unconditionally** ❌

## Change

**Commit 1 — skip the file.** Gate the explicit-file branch of `resolveArg` on `isMarkdown`, so an explicit path behaves exactly like the walk. Extension is the only signal available at resolve time (`.md`, `.markdown`, from the `mdpath` single-source-of-truth). All `check`/`fix`/`list`/`metrics` callers share this resolver, so the behavior is uniform.

**Commit 2 — warn instead of a silent no-op.** Skipping was correct but silent: `mdsmith fix .gitattributes` resolved to zero files and exited 0 with no output, so an explicit argument that did nothing looked like a clean pass. Added `ResolveOpts.OnSkipNonMarkdown`, a hook fired **only** in the explicit-file branch (never for walk/glob filtering, which stays silent by design). `check`/`fix` wire it to a stderr warning:

```
mdsmith: skipping ".gitattributes": not a Markdown file (expected .md, .markdown)
```

The warning is de-duplicated per path, gated to the **text** format, and suppressed under **`--quiet`** — because `check`/`fix` emit their diagnostics (including `--format json`/`sarif`) on stderr, so a prose line on that stream would corrupt structured output.

**MDS048 is unaffected.** The `.gitattributes` managed-block rule reads `.gitattributes` from disk itself (`bytelimit.ReadFileLimited`) while linting a *Markdown* file; the file never needs to be in the linted set.

## Behavior before / after

```console
# before
$ mdsmith check .gitattributes
stats: checked=1 fixed=0 failures=14 unfixed=14
$ mdsmith fix .gitattributes      # rewrites the file, still fails

# after (text)
$ mdsmith fix .gitattributes
mdsmith: skipping ".gitattributes": not a Markdown file (expected .md, .markdown)
$ echo $?      # 0, and the file is byte-for-byte unchanged

# after (machine-readable / quiet — warning suppressed, no stderr noise)
$ mdsmith check -f json .gitattributes   # no warning, clean stream
$ mdsmith check --quiet  .gitattributes  # no warning
```

A mixed run (`check README.md .gitattributes`) still lints `README.md`, drops the non-Markdown path, and warns about it.

## Tests

- **Unit** (`internal/lint/files_test.go`): explicit non-Markdown file skipped; mixed args keep only Markdown; `OnSkipNonMarkdown` fires for the explicit file only, never for a walked entry.
- **Unit** (`cmd/mdsmith/main_unit_test.go`): the warner emits + de-dupes in text mode; returns nil (suppressed) for quiet/json/sarif.
- **e2e** (`cmd/mdsmith/e2e_nonmarkdown_test.go`): `fix` leaves `.gitattributes` byte-for-byte unchanged and warns; `check` exits clean and warns; the warning is absent under `--quiet` and `-f json`; a mixed run reports the Markdown file and never anchors a diagnostic to the config file.

## Verification

- `go test ./...`, `go vet`, `golangci-lint` — all clean
- `mdsmith check .` — `checked=563 failures=0`
- Confirmed via probe that removing the guard makes all three e2e tests fail (real regression guards, not tautologies)

## Out of scope

The issue's **secondary observation** (an MDS048 scoped-vs-unscoped block disagreement) is explicitly marked not currently reproducible by the reporter. Not addressed here.

## Docs

`docs/reference/cli/check.md` and `fix.md` document that non-Markdown paths are skipped, that naming one explicitly prints a warning, and that `--quiet`/`json`/`sarif` suppress it.

https://claude.ai/code/session_011giDUosuDeqRDVWgmWyaj5

### DIFF
--- a/cmd/mdsmith/check.go
+++ b/cmd/mdsmith/check.go
@@ -113,6 +113,7 @@ func parseCheckFlags(args []string) (checkCLIOpts, []string, bool, int) {
 func checkFiles(fileArgs []string, opts checkCLIOpts) int {
 	cfg, cfgPath, logger, files, maxBytes, code := loadAndResolve(
 		fileArgs, opts.configPath, opts.verbose, opts.walk, opts.maxInputSize,
+		nonMarkdownSkipWarner(os.Stderr, opts.format, opts.quiet),
 	)
 	if code >= 0 {
 		return code
--- a/cmd/mdsmith/e2e_nonmarkdown_test.go
+++ b/cmd/mdsmith/e2e_nonmarkdown_test.go
@@ -33,8 +33,10 @@ func TestFix_ExplicitNonMarkdownFileUnchanged(t *testing.T) {
 	attrPath := filepath.Join(dir, ".gitattributes")
 	require.NoError(t, os.WriteFile(attrPath, []byte(gitattributesFixture), 0o644))
 
-	_, _, exitCode := runBinaryInDir(t, dir, "", "fix", ".gitattributes")
+	_, stderr, exitCode := runBinaryInDir(t, dir, "", "fix", ".gitattributes")
 	assert.Equal(t, 0, exitCode, "fix on a non-Markdown file should succeed with nothing to do")
+	assert.Contains(t, stderr, `skipping ".gitattributes"`,
+		"fix should warn that the explicitly named non-Markdown file was skipped, not do nothing silently")
 
 	got, err := os.ReadFile(attrPath)
 	require.NoError(t, err)
@@ -43,8 +45,9 @@ func TestFix_ExplicitNonMarkdownFileUnchanged(t *testing.T) {
 }
 
 // TestCheck_ExplicitNonMarkdownFileClean is the check-side counterpart:
-// a non-Markdown file named explicitly is skipped, so no Markdown
-// content diagnostics fire and the command exits clean.
+// a non-Markdown file named explicitly is skipped (no Markdown content
+// diagnostics, clean exit) but reported with a warning rather than a
+// silent no-op.
 func TestCheck_ExplicitNonMarkdownFileClean(t *testing.T) {
 	dir := t.TempDir()
 	isolateDir(t, dir)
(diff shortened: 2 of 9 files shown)


jeduden:
🟢 **Merge Queue** — picked up

**Next:** No action needed — you'll get another comment when CI starts on the batch. [View merge queue run](https://github.com/jeduden/mdsmith/actions/runs/30171123595).

jeduden:

Merged into batch branch [`merge-queue/batch-760-1785006745`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-760-1785006745) alongside #763, #765, #766. [View CI run](https://github.com/jeduden/mdsmith/actions/runs/30171197134).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`94af540`](https://github.com/jeduden/mdsmith/commit/94af5401979acae57c4a641621a302489eaee3b6). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/30171197134).

PR #758, opened 2026-07-20
feat(rules): add MDS060 occurrence rule (plan 2607022118)

jeduden:
## Summary

- Add MDS060 `occurrence` rule: counts tokens or a regex pattern per scope unit (paragraph, section, file) and flags when the count falls below `min` or exceeds `max`
- Fenced and indented code blocks are excluded via `CollectSectionParagraphsWithText`
- Opt-in rule (`enabled: false` by default); supports `lists:` wordlist integration via `WordlistTarget() = "tokens"`
- 32 unit tests + 5 integration fixture tests (3 bad, 2 good); all tests green

## Code review fixes applied

Four confirmed bugs found during xhigh code review and fixed before committing:

- **`File: f.Path` missing from diagnostics** — `diagEach`/`diagCombined` produced diagnostics with empty `File`, breaking CLI output filename display and sort order; all 61 other rules set `File: f.Path`
- **Allocation budget violation** — `countToken` called `strings.ToLower(text)` once per token per paragraph; with N tokens and P paragraphs that's N×P allocations. Fixed by pre-lowercasing text once per scope unit before the token loop, keeping `Check` within the ≤10-alloc budget
- **`FindAllString` allocates matched substrings** — switched to `FindAllStringIndex` which returns positions only, eliminating one string copy per match
- **`min < 0` accepted silently** — negative `min` made the lower-bound check always true; `ApplySettings` now returns an error for `min < 0`

## Files changed

- `internal/rules/occurrence/rule.go` — rule implementation
- `internal/rules/occurrence/rule_test.go` — 32 unit tests
- `internal/rules/MDS060-occurrence/README.md` — rule documentation
- `internal/rules/MDS060-occurrence/bad/` — 3 bad fixtures
- `internal/rules/MDS060-occurrence/good/` — 2 good fixtures
- `internal/rules/all/all.go` — import registration
- `internal/integration/perrule_bench_test.go` — alloc ceiling entry (MDS060: 4)
- `internal/integration/testdata/rule_walk_audit.json` — regenerated walk audit
- `internal/rulelayer/rule_walk_audit.json` — copy kept in sync
- `docs/research/markdownlint-coverage/README.md`, `internal/rules/index.md` — catalog regeneration

---
_Generated by [Claude Code](https://claude.ai/code/session_01FgvajK8jHk49ncLChABAq2)_

PR #756, opened 2026-07-20
refactor(schema): deduplicate isClaimed — export as schema.IsClaimed (plan 2607191918)

jeduden:
## Summary

- Exports `IsClaimed` from `internal/schema/validate.go` as the single implementation of the claimed-range set check.
- Removes the byte-for-byte private `isClaimed` copy from `internal/rules/requiredstructure/rule.go` (which already imported `internal/schema`), replacing the three call sites with `schema.IsClaimed`.
- Renames the private `isClaimed` → `IsClaimed` throughout the schema package (`validate.go`, `matchtree.go`, `validate_content.go`), eliminating the now-redundant forwarding wrapper.
- Adds `TestIsClaimed` in `internal/schema/validate_coverage_test.go` covering the present-index and absent-index branches per the project's Red/Green TDD convention.
- Trims the doc comment to one line, removing the caller enumeration and refactor-history narration that violated CLAUDE.md's no-caller-reference rule.
- Removes the double blank line left by the deletion in `requiredstructure/rule.go`.

Closes plan 2607191918.

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./internal/schema/... ./internal/rules/requiredstructure/...` passes
- [x] `go test ./...` passes (no regressions)
- [x] `mdsmith check .` passes (0 failures)
- [x] `TestIsClaimed` covers present-idx (true) and absent-idx (false) branches

---
_Generated by [Claude Code](https://claude.ai/code/session_01SZhZLDQbbu9MyHzgL1Sfjs)_

PR #754, opened 2026-07-19
test(init): add unit tests for printInitCatalog and setInitUsage (plan 2607191917)

jeduden:
## Summary

- Adds `TestPrintInitCatalog` directly exercising `printInitCatalog` via a `bytes.Buffer` — no subprocess, asserts both section headers and concrete description text for the `okf` starter and `wordlists` pack.
- Adds `TestSetInitUsage` directly exercising `setInitUsage` via a `bytes.Buffer` injected into the new `w io.Writer` parameter — no subprocess, asserts the static header flag names and pins `fs.PrintDefaults()` via the production `--force` description.
- Refactors `setInitUsage` to accept `io.Writer` (matching `printInitCatalog`) and calls `fs.SetOutput(w)` so both the header and `PrintDefaults()` output flow to the same writer without callers having to align them independently.

Closes plan 2607191917.

## Test plan

- [ ] `go test ./cmd/mdsmith/...` passes
- [ ] `TestPrintInitCatalog` and `TestSetInitUsage` appear in output with no subprocess
- [ ] `mdsmith check .` passes

---
_Generated by [Claude Code](https://claude.ai/code/session_013KgrSLVGA3ioVKvpNkK9ar)_

PR #762, opened 2026-07-21

jeduden:
## Summary

A codebase-wide audit against [`docs/development/high-performance-go.md`](docs/development/high-performance-go.md) was run using a fleet of agents scanning `internal/rules/**`, the core parse/lint pipeline (`internal/lint`, `internal/mdtext`, `internal/engine`, `pkg/markdown`, `pkg/goldmark`), and the schema/config/discovery surfaces. Most of the codebase turned out to already be heavily tuned against this exact doc (many files carry inline comments citing prior alloc-budget work), which narrowed the field considerably. Every candidate below was independently verified by reading the actual code and, where the claim was about CPU/allocation cost, by writing a real benchmark before touching anything — one high-severity candidate (a claimed `doublestar.Match` revalidation cost in `internal/discovery`) did not survive that check and was dropped.

The five fixes below are each backed by a red/green test that fails against the old code and passes against the new code, per this repo's TDD workflow. The PR then went through three rounds of adversarial xhigh-effort code review: round 1 found and fixed a real regression (see below), round 2 found and fixed a minor test-robustness issue, round 3 came back clean and recommended merge.

## Fixes

1. **`internal/lint`: closure-box allocation on `RunCache`/`File.Memo` warm path** — `File.Memo`, `File.MemoFile`, and `RunCache`'s shared `load()` helper each paid 2 allocations per cache-hit call: `sync.Map.LoadOrStore`'s second argument (a throwaway `&memoEntry{}`/`&runCacheEntry{}`) is constructed and discarded on every hit, and `sync.Once.Do(func(){ e.val = build() })` allocates its argument closure regardless of whether `Do` ends up invoking it. These are called at least once per host file across a workspace run. Fixed with a `Load`-before-`LoadOrStore` check plus the `atomic.Bool`+mutex double-checked-lock pattern (no wrapping closure) — verified at 0 allocs/op on the warm path (was 2).

2. **MDS021 (`include`): unconditional per-file alloc + mutex lock** — `Check`/`Fix` allocated a `visited` map and `chain` slice and took a rule-wide `sync.Mutex` (serializing this default-enabled rule's `Check` across the engine's parallel per-file fan-out) before ever checking whether the file has an `<?include?>` directive. Gated behind a cheap byte-needle pre-check, matching the `mayContainURL` pattern already used by `nobareurls`. (Round 1 of review caught that the first version of this gate only checked for `"<?include"`, silently dropping the "unexpected generated section end marker" diagnostic for a file with a dangling `<?/include?>` and no opening marker — fixed by checking both the start and end needles.)

3. **MDS027 (`cross-file-reference-integrity`): redundant per-file glob revalidation** — `Check` called `validateGlobSettings()` on every file, re-validating the same static `Include`/`Exclude` glob lists that `ApplySettings` already validated once at config-load time. Cached behind `cachedGlobSettingsErr` using the same atomic.Bool+mutex pattern as fix #1; `ApplySettings` resets the cache so reconfiguration (or a test that sets fields directly, bypassing `ApplySettings`) still gets a fresh, correct verdict.

4. **MDS043 (`no-reference-style`): ungated footnote regex scan** — `checkFootnotes` ran two full-file regex passes unconditionally on every `Check`, even for files with no footnote syntax at all. Benchmarked at ~585µs/op on 28KB of plain prose, ~0.5µs/op once gated behind a `mayContainFootnote` byte-needle check (both regexes require `"[^"` to match anything).

5. **`internal/foreignregion`: O(regions × lines) scan** — `Scan` walked the whole file once per configured region, re-converting every line via `strings.TrimSpace(string(line))` each time. Restructured into a single pass over `f.Lines` that evaluates every region's marker state against one shared `bytes.TrimSpace` per line, and switched to `[]byte`/`bytes.Equal` instead of allocating a `string` per line per region.

## Motivation

`mdsmith check` runs its rule set over every file in a workspace; the project's own performance doc states the budget philosophy plainly: "one extra alloc per `Check` means tens of thousands per run." Fixes #1–4 hit default-enabled or otherwise commonly-invoked code paths, so their savings apply broadly across a workspace check rather than to one rule in isolation.

## Test plan

- [x] Each fix has a dedicated test that fails on the pre-fix code and passes after (allocation-count assertions via `testing.AllocsPerRun`, or a benchmark with an inline budget for the one CPU-bound fix)
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l .` — clean
- [x] `go test ./...` (full suite, including `-race` on touched packages)
- [x] `go run ./cmd/mdsmith check .` — 563 files checked, 0 failures
- [x] Three rounds of adversarial xhigh-effort code review (1 real bug found and fixed, 1 minor test-robustness fix, 1 clean pass)
- [x] All 31 CI checks green

### DIFF
--- a/internal/foreignregion/scan.go
+++ b/internal/foreignregion/scan.go
@@ -9,6 +9,7 @@
 package foreignregion
 
 import (
+	"bytes"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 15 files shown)


jeduden:
🟢 **Merge Queue** — picked up

**Next:** No action needed — you'll get another comment when CI starts on the batch. [View merge queue run](https://github.com/jeduden/mdsmith/actions/runs/30170777831).

jeduden:

Merged into batch branch [`merge-queue/batch-754-1785006212`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-754-1785006212) alongside #754, #756, #758, #764. [View CI run](https://github.com/jeduden/mdsmith/actions/runs/30170910334).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`47ace27`](https://github.com/jeduden/mdsmith/commit/47ace276a7e86406831f87662468566f01861f3d). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/30170910334).

PR #770, opened 2026-07-25
perf: audit against high-performance-go.md, fix 5 measured hot paths

jeduden:
## Summary

Audited the Go core against `docs/development/high-performance-go.md` using a fleet of scanning agents (`internal/rules/**`, `internal/lint`/`internal/schema`/`pkg/goldmark`/`pkg/markdown`, and `cmd/mdsmith`/`internal/corpus`/`internal/linkgraph`/LSP) plus CPU/alloc profiling of `BenchmarkCheckCorpusLarge` (per the doc's own "profile before you rewrite" methodology).

Several agent-flagged leads did **not** survive measurement and were dropped rather than "fixed" for appearance's sake — each is a real anti-pattern shape, but not one that costs anything here:
- `internal/foreignregion/restore.go`'s `string(src[lineStart:lineEnd])` inside a `switch` — the Go compiler already elides this conversion's allocation for a switch/map-lookup comparison (measured 1 alloc for 53 lines, not 53).
- `doublestar.Match` vs `doublestar.MatchUnvalidated` for `requiredstructure`'s pre-validated path patterns — identical ns/op and allocs/op in a direct A/B benchmark.
- Hoisting `filepath.Dir` out of `crossfilereferenceintegrity.resolveTargetOSPath`'s per-link loop — the allocation is in `filepath.Join`'s result (distinct per link), not `Dir`, so hoisting doesn't reduce it.

Five issues did survive rigorous red/green verification (failing test/benchmark against the pre-fix code, passing after) and are fixed here, each its own commit:

1. **`astutil.HeadingText`/`HeadingTextBase` — memoize per heading node.** `no-trailing-punctuation`, `no-duplicate-headings`, `heading-increment`, and `first-line-heading` each independently re-walked the same heading's children and re-allocated its text via `bytes.Buffer.String` on every Check pass. Profiling showed `bytes.Buffer.String` at 28% of `BenchmarkCheckCorpusLarge`'s total allocations, 68% of that from `HeadingText`/`HeadingTextBase`. A first attempt routed the cache through the existing `sync.Map`-backed `Memo`/`MemoFile` facility and *regressed* allocations (sync.Map's per-insert bookkeeping costs more than a write-once, read-a-few-times keyset like headings saves); a plain mutex-guarded map fixed that.
2. **`schema.ExtractDocHeadings` — memoize per File.** `Validate`, `ValidateContent`, `ValidateAcronyms`, `BuildMatchTree`, and `requiredstructure.applyScopeRules` each independently re-walk `f.AST` for the same document's headings within one schema Check pass — `Validate` calls it directly and then calls `ValidateContent`, which calls it again. Now routed through `lint.File.MemoFile`, the same mechanism `astutil.CollectSectionParagraphs` already uses.
3. **`mdtext.Slugify` — pre-size the `strings.Builder`.** No `Grow()` call despite output length always being `<= len(s)`, so a non-trivial heading paid for several successive regrows. `Grow(len(s))` drops a ~50-byte heading from 5 allocations to 2. Runs once per heading from TOC generation, linkgraph, rename, and schema's index/cross-ref builders.
4. **`pkg/goldmark/text.blockReader.Value` — bulk-copy multi-line segments.** The inner copy loop appended one byte at a time; a single bulk `append(dst, src[i:end]...)` per line lets the compiler emit a memmove instead. ~205 ns/op → ~100 ns/op on a 3-line fixture (both 1 alloc/op — no allocation regression to catch here, so this is tracked as a plain benchmark rather than a `t.Fatalf`-gated one; see the commit for why).
5. **`engine.(*Runner).log` — share a disabled logger.** Allocated a fresh `&vlog.Logger{}` on every call when `-v` was off (the default) — one avoidable allocation per file in the workspace walk. `vlog.Logger.Printf` no-ops before touching its mutex when disabled, so a single shared instance is safe.

## Net effect

`benchstat` over 6 runs of `BenchmarkCheckCorpusLarge` (600-file corpus), origin/main vs this branch:

| metric | old | new | delta |
| --- | --- | --- | --- |
| allocs/op | 121.6k | 118.0k | **-2.98%** (p=0.002) |
| p95 | 103.5 ms | 97.0 ms | **-6.28%** (p=0.002) |
| us/file | 172.6 | 162.3 | **-5.97%** (p=0.002) |
| B/op | 70.65Mi | 70.68Mi | ~ (not significant) |

## Test plan

- [x] `go build ./...`
- [x] `go test -race` on every touched package (lint, engine, astutil, schema, goldmark/text)
- [x] `go vet ./...`
- [x] Each fix has a dedicated test/benchmark confirmed red against the pre-fix code, green after
- [x] `benchstat` comparison of `BenchmarkCheckCorpusLarge` before/after (table above)

Claude-Session: https://claude.ai/code/session_011EqgG5VSb7DYRvwakagT4B

---
_Generated by [Claude Code](https://claude.ai/code/session_011EqgG5VSb7DYRvwakagT4B)_

### DIFF
--- a/internal/engine/runner_log.go
+++ b/internal/engine/runner_log.go
@@ -9,13 +9,23 @@ import (
 	"github.com/jeduden/mdsmith/internal/rule"
 )
 
+// disabledLogger is the shared zero-value logger (*Runner).log
+// returns when no Logger is configured — the common case, since -v is
+// off by default. Every caller only reads Enabled or calls Printf,
+// which no-ops before touching the logger's mutex when Enabled is
+// false (internal/log.Logger.Printf), so nothing ever mutates this
+// instance; sharing it instead of allocating a fresh &vlog.Logger{}
+// on every call avoids one allocation per file in the workspace walk
+// (log runs once per file via Runner.checkFile).
+var disabledLogger = &vlog.Logger{}
+
 // log returns the runner's logger. If no logger is set, it returns a
 // disabled logger so callers don't need nil checks.
 func (r *Runner) log() *vlog.Logger {
 	if r.Logger != nil {
 		return r.Logger
 	}
-	return &vlog.Logger{}
+	return disabledLogger
 }
 
 // logRules logs each enabled rule in the effective config from the provided slice.
--- a/internal/engine/runner_log_test.go
+++ b/internal/engine/runner_log_test.go
@@ -1,0 +1,58 @@
+package engine
+
+import (
+	"testing"
+
+	vlog "github.com/jeduden/mdsmith/internal/log"
+)
+
+// TestRunner_log_NoLoggerAllocatesNothing pins that (*Runner).log
+// does not allocate when Logger is unset. log runs once per file in
+// the workspace walk (Runner.checkFile / lintFile), so a fresh
+// &vlog.Logger{} on every call — the common case, since -v is off by
+// default — is one avoidable allocation per file. A single shared
+// disabled-logger value is safe to return in its place: every caller
+// only reads Enabled or calls Printf, which no-ops before touching
+// the logger's mutex when Enabled is false, so nothing ever mutates
+// the shared instance.
+func TestRunner_log_NoLoggerAllocatesNothing(t *testing.T) {
+	r := &Runner{}
+
+	// Escapes to a package-visible sink so the compiler can't prove
+	// the result is unused and elide the allocation the way it would
+	// for a discarded `_ = r.log()` — real call sites (runner.go's
+	// `r.log().Printf(...)`, `sink := r.log()`) keep the pointer live
+	// past the call the same way.
+	sink := make([]*vlog.Logger, 0, 200)
+	allocs := testing.AllocsPerRun(200, func() {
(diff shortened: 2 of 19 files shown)


jeduden:
🔍 **Merge Queue** — bisecting

jeduden:
🔍 **Merge Queue** — bisecting

jeduden:
🔍 **Merge Queue** — bisecting

A larger batch failed CI. Bisection is isolating the culprit: this run tests up to **2 of 3** candidate PRs on [`merge-queue/batch-bisect-770-1786778789`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-bisect-770-1786778789). [View current bisect CI run](https://github.com/jeduden/mdsmith/actions/runs/31871965048).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`690c310`](https://github.com/jeduden/mdsmith/commit/690c3103b7cfda33b7c76f325aadfc069c89c5ce). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/31871965048).

PR #767, opened 2026-07-24
perf: audit against high-performance-go.md, fix 3 confirmed hot paths

jeduden:
## Summary

Audited the Go core against `docs/development/high-performance-go.md` using a fleet of scanning agents plus CPU/alloc profiling of `BenchmarkCheckCorpusLarge` (per the doc's own "profile before you rewrite" methodology). Most textbook anti-patterns the doc calls out (regex compiled per-call, missing pre-sized slices, redundant AST walks, `[]T{}` instead of `nil`, `ToLower`/`ToUpper` allocations) do **not** survive measurement in this codebase — it already went through a dedicated optimization pass (the "plan 195" / "plan 2606…" comments throughout `internal/`). Several promising-looking leads were implemented, benchmarked, and then **reverted** when they showed zero measurable improvement (e.g. converting `ast.Walk` closures to direct recursion — Go's escape analysis already keeps the walker off the heap here).

Three issues did survive rigorous red/green verification (failing benchmark/test against the pre-fix code, passing after) and are fixed here, ranked by measured impact:

1. **`no-reference-style` (MDS043) — missing byte-needle gate before two full-source regex scans.** `checkFootnotes` ran `footnoteRefRE`/`footnoteDefRE` over the entire source on every `Check`, even when the file has no footnote syntax at all. Added the same `bytes.Contains(source, "[^")` pre-check the doc's own example (`mayContainURL` in MDS012) already uses. **Measured: ~253,800 ns/op → ~4,500 ns/op (~56x) on a footnote-free fixture.**
2. **`build` (MDS039) — `strings.ToUpper` allocation on every path segment.** `hasReservedDeviceName` uppercased every segment before a set lookup; `ToUpper` only allocates when the input isn't already the target case, so real (lowercase) segments always paid for it. Every reserved name is 3–4 bytes, so a length pre-check skips the allocation for segments that can never match. **Measured: 602 ns/op & 10 allocs/op → ~380 ns/op & 5 allocs/op.**
3. **`git-hook-sync` (MDS048) — redundant `git rev-parse` subprocess.** `driftParts` resolved the hooks directory twice on a cache miss (once in `peekHookSource`, again in `preMergeCommitHookDrift`), each spawning a real OS subprocess — the exact "never exec a subprocess per item" anti-pattern the doc names. Resolved once and threaded through both helpers.

[...]

PR #765, opened 2026-07-23
perf: fix top 5 high-performance-go.md violations (struct layout + byte scans)

jeduden:
## Summary

An audit against `docs/development/high-performance-go.md` (four parallel scans covering allocation anti-patterns, the "patterns to avoid" table, struct layout/data structures, and strings/bytes/concurrency) surfaced several real violations. This PR fixes the top 5, ranked by hot-path frequency, safety, and testability:

1. **`internal/lsp.Diagnostic`** — `Range`/`Severity` (scalars) were declared before pointer-containing fields (`Code`, `Source`, `Message`, `Data`, `RelatedInformation`, `CodeDescription`). Go's GC computes a struct's ptrdata as the span from offset 0 through the last pointer-containing field, so those scalars sat inside the scanned span for nothing. This type's own doc comment calls it out as the "keystroke hot path" — built once per diagnostic on every `publishDiagnostics` call.
2. **`pkg/mdsmith.Diagnostic` / `RelatedLocation`** — same interleaving defect on the public Session API surface (WASM bindings, Obsidian plugin, any embedder), built once per diagnostic in `toDiagnostics`/`toRelatedLocations`. The sibling `internal/lint.Diagnostic` already carries this exact fix; these two were missed.
3. **`internal/output.jsonDiagnostic` / `jsonRelatedLocation`** — same defect on the CI / `--format json` output path, built once per diagnostic on every `mdsmith check --format json` run.
4. **`gensection.SplitLines`** — hand-rolled byte-by-byte scan for `'\n'` building the result via unsized `append`, instead of `bytes.Split`. This is a regression of a pattern the project already fixed once (`internal/lsp/diagnostics.go`'s `splitLines` uses `bytes.Split` for the same reason). Runs on every `Engine.Fix` pass for `<?include?>`/`<?catalog?>`-style directives. Measured: 6 allocs/op → 1 alloc/op on a representative 50-line body; 13 allocs/186µs → 1 alloc/43µs on a 2000-line stress input.
5. **LSP `documentEndPosition`** — manual per-byte newline-counting loop instead of `bytes.Count`, on every LSP "fix all" whole-document code action. Measured: ~63µs/op → ~1.6–2µs/op on a representative 2000-line document, allocations unchanged at zero.

[...]

PR #774, opened 2026-07-26
perf: audit against high-performance-go.md, fix 5 measured hot paths

jeduden:
## Summary

Audited the Go core against `docs/development/high-performance-go.md` using a fleet of scanning agents (`internal/rules/**`, `internal/lint`/`internal/mdtext`/`pkg/markdown`/`pkg/goldmark`/`internal/engine`, and `cmd/mdsmith`/`internal/lsp`/`internal/discovery`/`internal/index`) plus targeted before/after benchmarking of each candidate (per the doc's own "profile before you rewrite" methodology).

This repo has already been through several rounds of this exact audit (#762, #764, #765, #767), and — while scanning — I found #770 (branch `claude/kind-darwin-v8cn6a`) already open with a fifth, independent round covering `HeadingText`/`HeadingTextBase` memoization, `schema.ExtractDocHeadings` memoization, `Slugify` presizing, a goldmark bulk-copy fix, and a disabled-logger allocation. This PR deliberately does **not** touch any of that ground — all five fixes below are distinct hot paths #770 doesn't cover, so the two PRs should merge cleanly independent of order.

Five issues survived measurement and are fixed here, each its own commit with a dedicated before/after benchmark:

1. **`internal/discovery` — `matchesAny` re-validated already-validated patterns.** `validatePatterns` filters `w.patterns` to the valid set once up front, but `matchesAny` then called `doublestar.Match`, which re-runs `doublestar.ValidatePattern`'s parse walk on every file × pattern comparison during the workspace walk. `internal/globpath` already fixed this exact shape for the config-driven glob surfaces (its own comment cites "~4% of check CPU"); this mirrors it for file discovery, which runs on every `check`/`fix`/LSP-index invocation over every file in the tree. **`BenchmarkMatchesAny`: ~780 ns/op → ~685 ns/op (0 allocs/op both ways).**
2. **`cmd/mdsmith` — `orderFilesLeavesFirst` built its dependency index serially.** The loader (`bytelimit.ReadFileLimited` keyed by a read-only map, each call opening its own file handle) is safe for concurrent use, but the call used `BuildSerial` instead of the parallel `Build` that `internal/index` ships specifically for this shape (its own benchmarks show ~2x on a 1k-file corpus). Every `mdsmith fix` run paid the fully serial cost before the fix pass even starts. **`BenchmarkOrderFilesLeavesFirst` (500 files): ~30 ms → ~26 ms.**
3. **`catalog` (MDS-catalog) — numeric sort re-parsed every entry on every comparison.** `sortEntries`' numeric-mode comparator called `parseSortInt` (a CUE-path tokenize + map walk + `strconv.ParseInt`) on every comparison — O(n log n) re-parses — even though `allParseAsInt` already parses every entry once just to decide whether numeric mode applies. The parsed `int64` is now cached alongside the sort's existing lowercase-key precompute. **`BenchmarkSortEntriesNumeric` (5000 entries): ~23.4 ms / 307,798 allocs → ~3.9 ms / 29,806 allocs.**
4. **`pkg/goldmark/ast` — `BaseNode`'s `childCount` sat between pointer fields.** `BaseNode` declared a scalar (`childCount`) between its `Node` pointer fields and `attributes` ([]Attribute, pointer-ish), extending GC ptrdata over it for nothing. Every other struct-layout test in the package skips index 0 (the embedded base), so `BaseNode`'s own layout had never been checked — despite being embedded in every concrete AST node type parsed. No `unsafe` code depends on the field order (verified via grep).
5. **`astutil.SectionBody` scanned every paragraph on every call.** MDS057 (`required-text-patterns`) and MDS058 (`required-mentions`) each call `SectionBody` once per heading over the same paragraph slice, making a single `Check` O(headings × paragraphs). Both headings and paragraphs are already sorted by source line, so `sort.Search` now bounds the matching range directly instead of a full scan. **`BenchmarkSectionBodyPerHeading` (500 headings × 5 paragraphs): ~1.44 ms / 1500 allocs → ~0.12 ms / 1000 allocs.**

## Test plan

- [x] `go build ./...`
- [x] `go test ./...` (full suite green, 161 packages)
- [x] `go vet ./...`
- [x] `go test -race` on every touched package (discovery, cmd/mdsmith, catalog, goldmark/ast, astutil, index)
- [x] `internal/integration` per-rule alloc/timing budget gates and the rule-walk-audit manifest pass unchanged (MDS057/MDS058's budgets specifically re-verified after the `SectionBody` change)
- [x] Each fix has a dedicated benchmark confirmed slow against the pre-fix code, faster after (shown in each commit message)

Co-Authored-By: Claude Sonnet 5

---
_Generated by [Claude Code](https://claude.ai/code/session_01EepuKqu4XyH1XN5ZnaK3My)_

### DIFF
--- a/cmd/mdsmith/fix.go
+++ b/cmd/mdsmith/fix.go
@@ -321,8 +321,14 @@ func orderFilesLeavesFirst(files []string, rootDir string, maxBytes int64) []str
 		rels = append(rels, rel)
 	}
 
+	// Build, not BuildSerial: the loader only reads relToAbs (built
+	// above and never mutated again) and opens its own file handle per
+	// call via bytelimit.ReadFileLimited, so it is safe for the
+	// concurrent calls Build makes — see index.Index.Build's doc comment
+	// and internal/index's benchmarks for the general ~2x speedup this
+	// call site inherits on a workspace-sized file set.
 	idx := index.New(rootDir)
-	idx.BuildSerial(rels, func(rel string) ([]byte, error) {
+	idx.Build(rels, func(rel string) ([]byte, error) {
 		return bytelimit.ReadFileLimited(relToAbs[rel], maxBytes)
 	})
 
--- a/cmd/mdsmith/fix_order_bench_test.go
+++ b/cmd/mdsmith/fix_order_bench_test.go
@@ -1,0 +1,52 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/jeduden/mdsmith/internal/bytelimit"
+)
+
+// writeSyntheticFixCorpus writes n independent Markdown files (no
+// cross-links, so orderFilesLeavesFirst's dependency sort is a no-op
+// and every measured nanosecond goes to the index build's file reads)
+// under dir and returns their absolute paths.
+func writeSyntheticFixCorpus(t testing.TB, dir string, n int) []string {
+	t.Helper()
+	files := make([]string, 0, n)
+	for i := 0; i < n; i++ {
+		p := filepath.Join(dir, fmt.Sprintf("file_%05d.md", i))
+		body := fmt.Sprintf("# Heading %d\n\nSome body text for file %d.\n", i, i)
+		if err := os.WriteFile(p, []byte(body), 0o644); err != nil {
+			t.Fatalf("write %s: %v", p, err)
+		}
+		files = append(files, p)
+	}
+	return files
+}
+
+// BenchmarkOrderFilesLeavesFirst measures orderFilesLeavesFirst's cost
+// building the dependency index over a workspace-sized file set.
+// bytelimit.ReadFileLimited opens its own file handle per call and
+// touches no shared state, so the loader is safe for the concurrent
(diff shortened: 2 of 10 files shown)


jeduden:
🔍 **Merge Queue** — bisecting

jeduden:
🔍 **Merge Queue** — bisecting

jeduden:
🔍 **Merge Queue** — bisecting

A larger batch failed CI. Bisection is isolating the culprit: this run tests up to **2 of 3** candidate PRs on [`merge-queue/batch-bisect-770-1786778789`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-bisect-770-1786778789). [View current bisect CI run](https://github.com/jeduden/mdsmith/actions/runs/31871965048).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`690c310`](https://github.com/jeduden/mdsmith/commit/690c3103b7cfda33b7c76f325aadfc069c89c5ce). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/31871965048).

PR #775, opened 2026-07-27
perf: audit against high-performance-go.md, fix top 5 hot-path violations

jeduden:
## Summary

A fresh audit against [`docs/development/high-performance-go.md`](docs/development/high-performance-go.md), run after 13+ prior rounds of the same audit had already cleared most low-hanging fruit in `internal/rules/**`. Four parallel scans (allocation budget, string/byte patterns, data structures/struct layout, and skip-work/anti-patterns) turned up real, currently-present violations concentrated in `pkg/goldmark` (the vendored parser — excluded from `.golangci.yml` linting, so it had drifted from the linted tree) and `internal/metrics`. Each fix below is a separate red/green TDD commit: a failing test first (confirmed failing against the pre-fix code), then the minimal fix to make it pass.

## Fixes

1. **`pkg/goldmark/ast`: `BaseNode` field reorder** — `BaseNode` is embedded in every AST node (Document, Heading, Paragraph, Text, Emphasis, Link, …), making it the single most-allocated struct in the parser. `childCount` (a scalar) sat between the pointer-typed sibling/parent fields and `attributes` (a pointer-bearing slice), forcing GC ptrdata to scan past it for nothing. Reordered per "Group pointer fields first, scalars last."

2. **`internal/lint`: `lc0Pass` field reorder** — grouped its pointer-bearing fields ahead of its scalars for layout hygiene. (Reviewed down to: `lc0Pass` is actually stack-allocated and only reached via a default-off measurement seam today, so this is insurance against a future escape rather than a measured production win — comments were corrected to say so.)

3. **`pkg/goldmark/parser`: heading-ID collision loop** — `ids.Generate` built each disambiguation suffix with `fmt.Sprintf("%s-%d", result, i)` inside its collision loop. A heading colliding N deep (common in changelogs/release notes where a heading like "Changed" repeats every version) allocated N times just walking the loop. Now reuses one `[]byte` buffer via `strconv.AppendInt`.

4. **`pkg/goldmark/extension`: table delimiter alignments pre-size** — `parseDelimiter` already knows the exact column count from `bytes.Split` before its alignment loop, but grew `alignments` via a bare `append()` from nil, causing repeated reallocate-and-copy on wide tables. Now pre-sized with `make([]ast.Alignment, 0, len(cols))`.

[...]

PR #769, opened 2026-07-25
feat(mds072): SSRF guard and per-run egress ceiling

jeduden:
## Summary

Closes S001 (Medium) and S002 (Low) from the 2026-07-24 post-audit diff review ([plan/2607242010](plan/2607242010_mds072-ssrf-network-hardening.md)).

MDS072 previously probed document-supplied URLs with no IP filtering and followed redirects with a zero-value `http.Client`, creating an SSRF vector. It also capped concurrency but not the total number of distinct URLs probed per run. This PR adds defense in depth:

- **SSRF guard** (`ssrfControl` via `net.Dialer.Control`): refuses TCP connections to loopback, RFC 1918 private, link-local, ULA, CGN shared-address space, and cloud-metadata IPs (AWS/GCP/Azure `169.254.169.254`, Alibaba `100.100.100.200`) when `links.external-allow-internal` is false (the default).
- **Redirect guard** (`ssrfCheckRedirect` via `http.Client.CheckRedirect`): reasserts the IP guard on every redirect hop — both IP literals and DNS-resolved hostnames — before the new connection is dialled.
- **`links.external-allow-internal`** (bool, default `false`): opt-in escape hatch for users linting internal documentation from a trusted network position.
- **`links.external-max-probes`** (int, default `1000`): per-run probe ceiling via `atomic.Int64` counter. URLs past the cap are reported as `"external URL not probed: per-run limit reached"` rather than silently skipped, so truncated coverage is visible.
- **MDS072 README**: expanded settings table, new SSRF guard section (lists all blocked ranges), new egress ceiling section.
- **`docs/reference/telemetry.md`**: scoped the "zero outbound network calls" claim to the default configuration; added opt-in network access section for MDS072.

All `httptest` servers in the test suite bind to `127.0.0.1` (loopback). Test helpers now inject `external-allow-internal: true` by default so protocol-behaviour tests continue to reach the server. Dedicated SSRF tests use a separate helper that leaves the guard enabled.

## Test plan

- [ ] `go test ./...` — all packages green
- [ ] `go tool -modfile=tools/go.mod golangci-lint run` — clean
- [ ] `go run ./cmd/mdsmith check .` — docs pass
- [ ] `TestIsRestrictedIP_Blocked` covers all blocked ranges (loopback, private, link-local, ULA, CGN, metadata)
- [ ] `TestProbe_LoopbackBlocked` confirms a loopback httptest server is denied
[...]

PR #776, opened 2026-07-28
perf: audit against high-performance-go.md, fix 5 measured allocation/CPU hot spots

jeduden:
## Summary

Audited the Go core against `docs/development/high-performance-go.md` using a fleet of five parallel scanning agents (allocations, strings/bytes, data structures/struct layout, skip-work/inlining, concurrency/anti-patterns), then verified every candidate by hand before touching code — per the doc's own "profile before you rewrite... decide with benchstat, not eyeballs" discipline.

Several plausible-looking leads from the scan did **not** survive measurement and were dropped rather than shipped: a `fmt.Sprintf` in the vendored goldmark parser's heading-ID collision path turned out to be dead code (never enabled by mdsmith's own parser construction); a `[]byte(trimmed)` conversion in `internal/rules/include` compiled away to 0 allocations already (escape analysis); a `sync.RWMutex`→`Mutex` swap in the LSP server was dropped because genuine concurrent readers exist there (background lint vs. foreground hover/codeAction); and a couple of suggested struct field reorders (`pkg/markdown.Document`, `duplicatedcontent.corpusScanConfig`) turned out to save zero padding bytes when measured with `unsafe.Sizeof`. This mirrors the prior audit's own experience: "most textbook anti-patterns... do not survive measurement in this codebase."

Five issues did survive rigorous red/green verification and are fixed here, ranked by measured impact:

1. **MDS003 (`heading-increment`) + MDS005 (`no-duplicate-headings`) — redundant full-tree `ast.Walk`.** Both are default-on rules that each ran their own complete AST walk to collect headings. Added `astutil.CollectHeadingNodes`, memoized per `*lint.File` (mirroring the existing `CollectSectionHeadings`), and routed both rules through it so the second rule to run against a file hits the cache instead of re-walking. A manual 10-run benchstat comparison on a 201-heading document shows the shared walk cuts combined wall-clock by **~7% (101.9us → 94.5us mean)**, statistically significant (t≈3.37) despite a small allocs/op increase from `MemoFile`'s bookkeeping overhead — fewer full tree walks wins on net. A deterministic test pre-seeds the memo with an empty result and asserts both rules read through it rather than re-walking `f.AST` directly (a timing-based test would have been confounded by an unrelated pre-existing per-file memo).
2. **`markdownflavor.fixGitHubAlerts` / `buildAlertSkipMaps` — unsized append and a string conversion that escapes.** `fixGitHubAlerts`'s output loop appended to a nil slice with a knowable bound (`len(f.Lines)`); pre-sized it, mirroring `codeblockstyle.Fix`'s identical pattern in the same rule family. `buildAlertSkipMaps` converted every lazy-continuation line to a `string` just to `TrimLeft` whitespace and check a `>` prefix — both doable directly on the `[]byte` line — and that conversion escaped (its result crossed a `strings.HasPrefix` call), so every continuation line paid a full copy. **Measured: `fixGitHubAlerts` 169 → 113 allocs/op; `buildAlertSkipMaps` alone 60 → 10 allocs/op** on a 50-alert fixture.
3. **`catalog.extractPlaceholderFields` — unsized append.** Its dedup companion map was already pre-sized to `len(all)`; the output slice wasn't. Pre-sized it, keeping the project's "return nil, not `[]T{}`" convention for the no-placeholders case. **Measured: 13 → 11 allocs/op.**

Each fix has a dedicated regression test with a pinned budget, verified to fail against the pre-fix code (red) and pass after (green) via `git stash`/`git stash pop` round-trips during development.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] Each fix's new test confirmed red against the pre-fix code (`git stash`), green after
- [x] `internal/integration/testdata/rule_walk_audit.json` regenerated via `MDSMITH_REGEN_WALK_AUDIT=1` (MDS003/MDS005's `uses_ast_walk` signal correctly flips to `false`) and the embedded copy in `internal/rulelayer/` re-synced

Claude-Session: https://claude.ai/code/session_018nw8T7sXArBJGuwEM7ciif

---
_Generated by [Claude Code](https://claude.ai/code/session_018nw8T7sXArBJGuwEM7ciif)_

### DIFF
--- a/internal/integration/testdata/rule_walk_audit.json
+++ b/internal/integration/testdata/rule_walk_audit.json
@@ -29,7 +29,7 @@
     "code_block_sensitive": false,
     "fired": true,
     "is_node_checker": false,
-    "uses_ast_walk": true,
+    "uses_ast_walk": false,
     "reads_file_ast": true
   },
   {
@@ -51,7 +51,7 @@
     "code_block_sensitive": false,
     "fired": true,
     "is_node_checker": false,
-    "uses_ast_walk": true,
+    "uses_ast_walk": false,
     "reads_file_ast": true
   },
   {
--- a/internal/rulelayer/rule_walk_audit.json
+++ b/internal/rulelayer/rule_walk_audit.json
@@ -29,7 +29,7 @@
     "code_block_sensitive": false,
     "fired": true,
     "is_node_checker": false,
-    "uses_ast_walk": true,
+    "uses_ast_walk": false,
     "reads_file_ast": true
   },
   {
@@ -51,7 +51,7 @@
     "code_block_sensitive": false,
     "fired": true,
     "is_node_checker": false,
-    "uses_ast_walk": true,
+    "uses_ast_walk": false,
     "reads_file_ast": true
   },
   {
--- a/internal/rules/astutil/astutil.go
+++ b/internal/rules/astutil/astutil.go
@@ -110,6 +110,40 @@ func buildSectionHeadings(f *lint.File) any {
 	return out
 }
 
+// CollectHeadingNodes returns every *ast.Heading node in the document,
+// in source order. Unlike CollectSectionHeadings (which only exposes
+// Level and Line), this keeps the node itself so a caller can extract
+// heading text via HeadingText.
+//
+// Memoized per File via lint.File.MemoFile: MDS003 (heading-increment)
+// and MDS005 (no-duplicate-headings) both previously ran their own
+// full ast.Walk to collect the same headings; sharing one walk here
+// means the second rule to run pays a cache hit instead of re-walking
+// the tree (docs/development/high-performance-go.md, "memoize
+// per-input computations").
+func CollectHeadingNodes(f *lint.File) []*ast.Heading {
+	return f.MemoFile("astutil.headingNodes", buildHeadingNodes).([]*ast.Heading)
+}
+
+// buildHeadingNodes is the MemoFile-style builder for the heading-nodes
+// memo. Defined at package scope so the value passed to MemoFile is a
+// plain function pointer, matching buildSectionHeadings.
+func buildHeadingNodes(f *lint.File) any {
+	var out []*ast.Heading
+	_ = ast.Walk(f.AST, func(n ast.Node, entering bool) (ast.WalkStatus, error) {
+		if !entering {
+			return ast.WalkContinue, nil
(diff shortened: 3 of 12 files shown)


jeduden:

jeduden:
The `queue:failed` merge-conflict here isn't a real git conflict against `main` — this branch already contains all of `main` (`git merge-base --is-ancestor origin/main HEAD` succeeds cleanly, nothing to rebase). CI on the latest commit is fully green.

The conflict is against sibling PRs: there are currently 7 other open PRs actively in the queue (`#785`, `#786`, `#788`, `#790`, `#792`, `#793`, `#794`) plus several more open-but-unqueued (`#769`, `#770`, `#774`, `#775`, `#779`, `#780`, `#782`, `#783`) that are all near-duplicates of this one — repeated runs of the same scheduled "audit against `high-performance-go.md`" task, independently converging on overlapping fixes (heading-walk sharing, `markdownflavor` allocation fixes, `catalog` pre-sizing all appear in more than one of these). Re-adding the `queue` label right now would just add an 8th PR to a pile that's already failing to batch-merge itself.

Holding off on re-queuing until the duplicate PRs are triaged — flagged to the user for a decision on which one to land and which to close, rather than guessing unilaterally.

---

jeduden:
Superseded by #800, which consolidates this PR with #779, #790 and #793 and resolves every conflict by hand.

Carried forward from this PR:

- `astutil.CollectHeadingNodes` + `HeadingNodesMemoKey` — the shared memoized heading walk between MDS003 and MDS005
- `noduplicateheadings` switched to that shared walk (kept with the direct `HeadingText` call, exactly as shipped here — pairing it with `main`'s `HeadingTextCached` pushed MDS005 to 12 allocs/op against CLAUDE.md's ceiling of 10, since each helper installs its own per-File memo)
- `headingincrement` and `catalog` changes, and the `rule_walk_audit.json` updates
- Both alloc tests in `markdownflavor/alloc_test.go`, unioned with `main`'s. Their budgets were re-run against `main`'s implementation and still hold.

Not carried forward:

- `markdownflavor/rule.go`: `main` (via #783) landed the same two fixes with better implementations — `f.LineOfOffset` rather than `flavor.LineCol`, and a `bytes.Buffer` rather than a pre-sized `[]string` + `strings.Join`.

PR #779, opened 2026-07-29
perf: fix top hot-path violations of high-performance-go.md

jeduden:
## Summary

A follow-up audit against `docs/development/high-performance-go.md`, in the same spirit as PR #767. A team of scanner agents swept `internal/rules/**`, `internal/lint`, `internal/mdtext`, `internal/discovery`, `pkg/markdown`, and `pkg/goldmark` for violations of the guideline's patterns (regex/string handling, allocation/struct layout, and skip-work/memoization), excluding the packages #767 already fixed (`build`, `githooksync`, `noreferencestyle`). Findings were ranked by whether the rule/path runs by default and how often it fires on a real workspace walk, then the top 5 were fixed with red/green TDD and verified with `benchstat` per the doc's own "Decide with benchstat, not eyeballs" methodology.

**Three rounds of xhigh code review ran against this PR.** Rounds 1 and 2 found and fixed real correctness bugs, both rooted in the same false assumption (`doublestar.ValidatePattern` returning true implies `Match` and `MatchUnvalidated` agree) — see items 3 and (1, 2) below. Round 3 found no correctness issues and additionally *proved* (not just fuzzed) that the fix's remaining safety guard is sound. Every claim in this PR, including each review round's own claims, was reproduced independently before being trusted.

## Fixes

1. **`internal/discovery/discovery.go`** — `walker.matchesAny` called `doublestar.Match` per configured pattern per file during every `check`/`fix`/`backlinks`/`deps`/`rename` workspace walk. Switched to `doublestar.MatchUnvalidated` for patterns without brace (`{...}`) syntax, falling back to safe `Match` for the rare pattern that uses it (see the divergence writeup below). **-6.65% (p=0.000, n=10)** on the realistic mismatch-heavy case, guard cost is noise-level.
2. **`internal/rules/requiredstructure/rule.go`** — `checkPathPatterns` (MDS020, default-enabled) re-validated a `path-pattern:` glob already validated once at config-parse time, on every file. Same guarded `MatchUnvalidated`/`Match` fix. No regression at the integration-benchmark level (p=0.075) — the underlying per-call win is real but swamped by diagnostic-construction cost in an all-mismatch benchmark shape.
3. **`internal/rules/catalog/rule.go`** — `isExcluded` re-validated each `exclude:` pattern per glob-matched candidate. Initially swapped to `MatchUnvalidated` like the two fixes above (**-13.50%, p=0.000, n=10**), but round-1 review found this call site is *not* sound like the other two: `checkInjection`/`checkCaseMismatches` reach `isExcluded` with a directive's raw, **unvalidated** params, bypassing the engine's `Validate` pass entirely. A cached-validity guard restores correctness but `benchstat` then shows no measurable difference from plain `Match` (p=0.669) — so `isExcluded` reverts to `Match` outright rather than carrying extra complexity for zero measured benefit. `TestIsExcluded_MalformedPatternNeverMatches` pins the invariant permanently; round-3 confirmed the net diff against `main` is a true no-op (byte-identical function body).
4. **`pkg/goldmark/ast/ast.go`** — `BaseNode` (embedded in every AST node type, instantiated once per node on every parse of every file) declared a scalar field (`childCount`) between its pointer fields and `attributes []Attribute`, extending GC ptrdata past a scalar for no reason — exactly the anti-pattern the project's own `fieldorder.AssertPointersBeforeScalars` check exists to catch, just never applied to `BaseNode` itself. Reordered; struct size unchanged (verified via `unsafe.Sizeof`, and independently by summing field sizes), only the scanned ptrdata prefix shrinks (96→88 bytes). New `TestBaseNode_PointerFieldsBeforeScalars` pins it.
5. **`internal/rules/toc/rule.go`** — `escapeLinkText` ran three chained `strings.ReplaceAll` scans (once per TOC item, on every `Check()` for a file with `<?toc?>`, not only `Fix()`). Replaced with a `strings.IndexByte`-guarded fast path plus a single `strings.Builder` pass. **-56.39% / -48.97%** time on the no-escape/escape cases (p=0.000, n=10), allocs **3 → 1** when escaping is needed. An earlier version of this fix used `strings.ContainsAny` as the guard and actually *regressed* the common case by +53.8% — caught by `benchstat` before landing.

### The `ValidatePattern` ≠ `Match`-agrees-with-`MatchUnvalidated` finding

In vendored doublestar v4.10.0, a brace group tries each alternative via recursive calls and short-circuits with `if result || err != nil { return result, err }` — so a syntax-error-triggering construct in a *non-final* alternative aborts the whole match before a later, perfectly matching alternative ever runs. `MatchUnvalidated` skips the validation step that raises that error, so it keeps trying alternatives and can return `true` where `Match` returns `false` with an error — even for a pattern `ValidatePattern` itself accepts. Confirmed directly: `doublestar.Match("{[!mdb[],docs/**/*.md}", "docs/a.md")` → `(false, error)`, `MatchUnvalidated` → `true`.

Round 3 went further and proved this is a *complete* characterization, not an empirically-lucky one: reading the matcher's control flow shows `validate` can only change a match's outcome through that same brace-recursion short-circuit, and pattern rewriting only ever happens inside that same `case '{'` branch — so a pattern with no `{` byte can *never* diverge, by construction, not just "in every case tried so far." Items 1 and 2 above use a cheap `strings.IndexByte(p, '{') < 0` guard to take the fast path only when it's provably safe; a `\{` escaped brace harmlessly routes to the slow path even though it isn't real brace syntax (a false-positive, not a correctness gap).

## Test plan

- [x] Every fix has a dedicated red/green test, confirmed red against the pre-fix code and green after: `escapeLinkText` unit tests, `isExcluded` malformed-pattern guard, `BaseNode` struct-layout test, and end-to-end brace-alternate regression tests (both the divergence case and the ordinary-match success case) for both `discovery.Discover()` and `requiredstructure.Rule.Check()`.
- [x] `go build ./...`
- [x] `go test ./...` (full suite, all packages green) and `go test -race` on every touched package
- [x] `go vet ./...`
- [x] `go tool -modfile=tools/go.mod golangci-lint run` — full repo, 0 issues
- [x] `go test -tags goldmark_upstream ./pkg/goldmark/...` (the fork A/B parity job)
- [x] Every perf change backed by `benchstat` (`-count=10`) before/after comparison, documented in the commit messages
- [x] Three independent rounds of xhigh code review completed; every correctness finding was reproduced independently (against the vendored `doublestar` source, via throwaway differential-fuzz and exhaustive-sweep tests) before being fixed, not taken on faith
- [x] Codecov patch coverage green (100% statement coverage on both guarded functions, including the success-branch tests a codecov failure surfaced were missing)
- [x] CI: 31/31 checks green

Claude-Session: https://claude.ai/code/session_01PuSczC6T2oegVQKNCeRqJx


jeduden:

jeduden:
Checked this against `main` directly (`git merge-tree` with the merge-base, main, and this branch's head): the merge is clean, zero conflict markers. The queue failure is against another PR in the same batch, not against `main` — this repository currently has ~15 open PRs (774, 775, 776, 780, 782, 783, 785, 786, 788, 790, 792, 793, 794, and others) all independently auditing against `docs/development/high-performance-go.md` and touching the same files (`internal/discovery/discovery.go`, `internal/rules/catalog/rule.go`, `internal/rules/requiredstructure/rule.go`, etc.), which is almost certainly why the batch can't merge cleanly.

Re-added the `queue` label since there's nothing to fix on this branch. Flagging the broader duplication separately.

---

jeduden:
Superseded by #800, which consolidates this PR with #776, #790 and #793 and resolves every conflict by hand.

The most important thing this PR contributed is carried forward in full: the **brace guard in `discovery.matchesAny`**.

`main` currently calls `doublestar.MatchUnvalidated` unconditionally. As documented here, `Match` and `MatchUnvalidated` provably diverge on a brace alternative whose non-final alternative contains a syntax error. This PR keeps the `MatchUnvalidated` fast path for brace-free patterns and falls back to validated `Match` when the pattern contains `{`. That is a correctness fix on top of the optimization, not a competing implementation, so #800 takes this side over `main`'s — and over #790's, which has the unguarded version.

Also carried forward: the `catalog`, `toc` (single-pass `escapeLinkText`), and `requiredstructure` changes with their tests and benchmarks.

Not carried forward:

- `BenchmarkWalker_matchesAny` — duplicates `main`'s `BenchmarkMatchesAny`, which is more faithful because it builds `w.patterns` through `validatePatterns`, as `Discover` does.
- `TestBaseNode_PointerFieldsBeforeScalars` — `main` already has it in `basenode_structlayout_test.go`; keeping both would not compile.

PR #780, opened 2026-07-30
perf: fix top 5 hot-path issues from a high-performance-go.md audit

jeduden:
## Summary

An audit against [`docs/development/high-performance-go.md`](https://github.com/jeduden/mdsmith/blob/main/docs/development/high-performance-go.md) fanned four parallel scans across `internal/rules`, the core engine, the parsing/text packages, and the CLI/LSP/misc packages, surfacing ~50 candidate violations of the guide's allocation, byte-vs-string, and pre-check patterns. This PR fixes the top 5, each with a red/green `testing.AllocsPerRun` (or call-count) regression test written before the fix, then hardened through three rounds of adversarial xhigh-severity code review before merge.

### The 5 fixes

1. **`internal/rename/rename.go`** — `validRefDefMatches` ran a full CommonMark parse on every workspace file on every heading-rename LSP request, even though the ref-def regex can only match a line containing `]:`. Gates the parse behind a cheap `bytes.Contains(body, []byte("]:"))` pre-check — the "gate expensive analyzers behind a cheap pre-check" pattern the guide names. **190 → 0 allocs/op** on a no-ref-def body.
2. **`internal/foreignregion/restore.go`** — `matchedRegionSpans` allocated a new string per line (`strings.TrimSpace(string(...))`) on every `Restore` call, which runs on every `mdsmith fix` pass when foreign regions are configured. Switched to `bytes.TrimSpace`/`bytes.Equal`, staying in `[]byte`. **603 → 3 allocs/op** on a ~450-line source.
3. **`internal/lsp/server_diagnostics.go`** — `surfaceForeignDiagnostics` called `logger.Printf` + `fmt.Sprintf` + a JSON-RPC notification once per diagnostic, so every lint pass (every keystroke) re-logged and re-notified every config-target finding individually — the guide's named anti-pattern "`log.Printf` per item in a hot loop". Now batches by severity (one `Printf`/notification per group) and builds each line with `strconv.Itoa`/`WriteString` instead of `fmt.Fprintf`.
4. **`internal/checker/checker.go`** — `CheckConfiguredRules`, the hottest per-file function in the codebase, grew its diagnostics slice via unsized `append` instead of pre-sizing from the already-known total.
5. **`pkg/goldmark/text/segment.go`** — `Segments.Value`, on the hottest per-node parse path, grew its output buffer via unsized `append` even though each `Segment.Len()` makes the total size computable in one pass. **6 → 1 alloc/op** on a 24-segment, 1200-byte collection.

### Review history

Three rounds of independent xhigh-effort adversarial review ran against the diff before merge, each followed by fixes (all in their own commits, referencing the round that found them):

- **Round 1** (5 issues, all fixed): `checker.go`'s pre-sized slice forced a non-nil empty result instead of `nil` (violates CLAUDE.md's "return nil, not `[]T{}`"); the checker alloc-budget test had only a 1-allocation red/green margin and flaked under `-race`; `segment.go`'s `ForceNewline` byte wasn't budgeted, defeating the fix for the common fenced/indented-code-block-at-EOF case; the LSP per-severity `Grow()` calls were structured wrong (called once per diagnostic on an empty builder, reserving nothing useful); a stale test comment.
- **Round 2** (5 issues, all fixed): `segment.go`'s own `ForceNewline` fix reintroduced the same nil-vs-empty bug round 1 had just fixed elsewhere; `checker.go`'s guard was simplified since it always reduced to `total > 0` in that context; the LSP `Grow` fix had zero test coverage; `fmt.Fprintf` — not the broken `Grow` — was the dominant allocation cost in the LSP diagnostic-line builder, replaced with `strconv`/`WriteString`; a note that the deprecated `ast.HTMLBlock.Text` now always reallocates on its trailing append (left as-is, off mdsmith's hot path).
- **Round 3** (3 issues, all fixed): a *third* instance of the nil-vs-empty bug — a zero-length `ForceNewline` segment (e.g. an empty fenced code block) still returned `[]byte{}` instead of `nil`, since the fix's total-based fast path couldn't see it; the round-2 LSP alloc test was missing the `raceEnabled` skip round 1 established and flaked under `-race`; a no-op `strings.Repeat("x", 1)` in a test fixture.

Round 3 confirmed the PR holds up under a fresh, from-scratch adversarial pass: byte-for-byte differential testing of the `fmt.Fprintf` → `strconv` rewrite (9 adversarial cases, 0 mismatches), 20,000-case differential fuzzing of the `foreignregion` byte-native rewrite, and re-verified every fix's red/green gate by reverting it individually.

No behavior changes beyond the intentionally batched LSP log/notification format. Every existing test in the touched packages passes unchanged.

## Test plan

- [x] `go build ./...` — clean
- [x] `go vet ./...` — clean
- [x] `go test ./... -count=1` — full suite passes
- [x] `go test -race` on all 5 touched packages — passes
- [x] `go run ./cmd/mdsmith check .` — passes (567 files, 0 failures)
- [x] Each of the 5 fixes has a red/green regression test; each of the 13 review-round fixes was independently reverted and reconfirmed red before being restored

https://claude.ai/code/session_017SdNJqyivvvVq3idsQKh4n

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -143,12 +143,12 @@ func CheckConfiguredRules(
 		runBlockCheckers(f, blockCheckers)
 	}
 
-	var diags []lint.Diagnostic
 	total := 0
 	for _, s := range slots {
 		total += len(s.diags)
 	}
-	if cap(diags) < total {
+	var diags []lint.Diagnostic
+	if total > 0 {
 		diags = make([]lint.Diagnostic, 0, total)
 	}
 	for _, s := range slots {
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -143,7 +143,11 @@ func CheckConfiguredRules(
 		runBlockCheckers(f, blockCheckers)
 	}
 
-	var diags []lint.Diagnostic
+	total := 0
+	for _, s := range slots {
+		total += len(s.diags)
+	}
+	diags := make([]lint.Diagnostic, 0, total)
 	for _, s := range slots {
 		diags = append(diags, s.diags...)
 	}
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -143,11 +143,14 @@ func CheckConfiguredRules(
 		runBlockCheckers(f, blockCheckers)
 	}
 
+	var diags []lint.Diagnostic
 	total := 0
 	for _, s := range slots {
 		total += len(s.diags)
 	}
-	diags := make([]lint.Diagnostic, 0, total)
+	if cap(diags) < total {
+		diags = make([]lint.Diagnostic, 0, total)
+	}
 	for _, s := range slots {
 		diags = append(diags, s.diags...)
 	}
--- a/internal/checker/perf_test.go
+++ b/internal/checker/perf_test.go
@@ -7,6 +7,7 @@ import (
 	"github.com/jeduden/mdsmith/internal/config"
 	"github.com/jeduden/mdsmith/internal/lint"
 	"github.com/jeduden/mdsmith/internal/rule"
+	"github.com/stretchr/testify/require"
 )
 
 // perfStubRule is a plain rule.Rule (not a NodeChecker/BlockChecker) whose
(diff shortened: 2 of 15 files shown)


jeduden:
🔍 **Merge Queue** — bisecting

jeduden:
🔍 **Merge Queue** — bisecting

jeduden:
⏳ **Merge Queue** — requeued

[View merge queue run](https://github.com/jeduden/mdsmith/actions/runs/31871651962).

jeduden:

Merged into batch branch [`merge-queue/batch-775-1786780171`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-775-1786780171) alongside #775, #782, #785, #786. [View CI run](https://github.com/jeduden/mdsmith/actions/runs/31872985040).

jeduden:
✅ **Merge Queue** — merged

PR #782, opened 2026-07-31

jeduden:
## Summary

Audited the codebase against [`docs/development/high-performance-go.md`](docs/development/high-performance-go.md) using four parallel research agents, each sweeping a different category of anti-pattern from the guideline (non-package-scope regex compiles, `fmt.Sprintf`/`string()` conversions on hot paths, un-presized slices and `[]T{}`-vs-`nil` returns, and the misc "patterns to avoid" table — `defer` in loops, `time.Now()` in loops, `log.Printf` per item, unbounded `os.ReadFile`, `context.Background()` deep in a call chain).

The codebase turned out to already be quite disciplined (four prior PRs — #762, #764, #765, #767 — already swept the obvious violations), so this pass found a small number of genuine, verifiable issues rather than a long tail. Two additional candidates the agents initially flagged (a `string()` conversion in `horizontalrulestyle.checkHR`, and a `[]string{}`-vs-`nil` return in `concisenessscoring.extractFeatures`) were investigated and **ruled out** after measurement — the guideline's own process ("profile before you rewrite — a flat profile in your suspected hot path stops a useless fix") caught them before they became no-op changes: the first is proven not to escape to the heap by the compiler's own escape analysis (`go build -gcflags="-m=2"` confirms "does not escape"), and the second is intentionally non-nil per an existing test (`TestClassify_EmptyInputKeepsCueSliceNonNil`).

### Fixes (each its own commit, red/green TDD)

1. **`pkg/mdsmith`: bound `frontMatterFor`'s read at the session's max-input-size.** `Session.Kinds` was the one `Workspace` read path with no size cap at all — every other read (`Check`/`Fix`) already threads `MaxInputBytes` through the engine runner. Added a `readFileLimited` capability to `OSWorkspace` and `OverlayWorkspace` (bytelimit-backed, so an oversized file is never read past `max+1` bytes) instead of `os.ReadFile`'s unbounded read.
2. **MDS054 (`no-undefined-reference-labels`): avoid a `string()` copy per bracket candidate.** When a `placeholders:` vocabulary is configured, three scanners each converted every reference-style bracket candidate to a string just to check it against the vocabulary — on a rule that's enabled by default. Replaced with a documented zero-copy `unsafe.String` view.
3. **MDS034 (`markdown-flavor`): avoid per-line allocation in `fixGitHubAlerts`.** The `Fix` path built its result in an un-presized `[]string` joined via `strings.Join`, and its skip-map builder converted every alert-paragraph line to a string just to test its first byte. Rewrote around a single pre-grown `bytes.Buffer` (mirroring `blockquotewhitespace.Fix`, the sibling rule that already made this switch) plus `bytes.TrimLeft`.
4. **MDS062 (`link-validity`): presize `reversedInLine`'s match slice.** The regex submatch-index slice already gives the exact upper bound before the result loop starts; presized instead of growing via bare `append`.
5. **MDS034: use the memoized line index instead of a per-call rescan.** `buildAlertSkipMaps` called `flavor.LineCol` (a fresh `bytes.Count`/`bytes.LastIndexByte` scan from byte 0) once per continuation line in a loop — O(n) per line, O(n²) over a long alert paragraph — instead of `lint.File`'s existing O(log n) memoized `LineOfOffset`. This is a pure CPU fix (allocations unchanged), so it's validated via `benchstat` rather than an alloc-budget test: **−52.00% time (p=0.000, n=10)** on a 500-line fixture.

## Test plan

- [x] `go build ./...`
- [x] `go test ./...` (161/161 packages pass)
- [x] `go vet ./...`
- [x] Each fix has a dedicated alloc-budget (or, for the CPU-only fix, benchmark/benchstat) test that fails before the change and passes after — confirmed by temporarily reverting each fix and re-running its test.

---
_Generated by [Claude Code](https://claude.ai/code/session_0143ESqh4f4KhojH5myCBRJv)_

### DIFF
--- a/docs/background/concepts/engine-api.md
+++ b/docs/background/concepts/engine-api.md
@@ -129,6 +129,11 @@ a `Root`, `ReadFile` resolves a relative URI against it, exactly as the
 different files. An absolute path is read unchanged, and an empty `Root`
 reads paths as passed. `MemWorkspace` keys both paths off the same map.
 
+`Session.Kinds` caps its front-matter read at the session's
+`max-input-size`, the same cap `Check`/`Fix` already apply to the
+engine's parse input. A file over the cap resolves the same lenient way
+a missing file does: no error, and no front-matter-derived kinds.
+
 A third implementation, `OverlayWorkspace`, is the LSP server's. It
 reads disk rooted at `Root` but lets `Set(uri, bytes)` shadow a path's
 content with an editor's unsaved buffer, so cross-file rules read the
--- a/internal/bytelimit/bytelimit.go
+++ b/internal/bytelimit/bytelimit.go
@@ -4,6 +4,7 @@
 package bytelimit
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"io/fs"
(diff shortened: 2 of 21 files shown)


jeduden:
🔍 **Merge Queue** — bisecting

jeduden:
🔍 **Merge Queue** — bisecting

jeduden:
⏳ **Merge Queue** — requeued

[View merge queue run](https://github.com/jeduden/mdsmith/actions/runs/31871651962).

jeduden:

Merged into batch branch [`merge-queue/batch-775-1786780171`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-775-1786780171) alongside #775, #780, #785, #786. [View CI run](https://github.com/jeduden/mdsmith/actions/runs/31872985040).

jeduden:
✅ **Merge Queue** — merged

PR #783, opened 2026-08-02
perf: fix top 5 high-performance-go.md violations from a 3-agent audit

jeduden:
## Summary

Three parallel subagents audited origin/main against
[`docs/development/high-performance-go.md`](docs/development/high-performance-go.md),
each covering a distinct slice of the guide (allocations/strings-bytes,
struct-layout/skip-work, concurrency/anti-pattern-table). This is the fifth
round of this recurring audit (after #762, #764, #765, #767), so most
low-hanging fruit was already gone; the findings below are what survived
verification against the current code.

Ranked by severity/impact, the top 5 real, safely-fixable issues found:

1. **`markdownflavor` (MDS034) `fixGitHubAlerts` / `buildAlertSkipMaps`** —
   both converted every scanned line to a `string` (one via an unsized,
   growing `[]string` accumulator, the other via a per-continuation-line
   `strings.TrimLeft(string(...))`). Both now stay in `[]byte`, and
   `fixGitHubAlerts` pre-sizes its accumulator. **176 → 70 allocs** on an
   addPrefix-heavy 50-alert fixture (measured against an `origin/main`
   worktree of the same fixture, not the original, less representative
   fixture the first commit shipped with — see the review-round commits for
   that correction).
2. **`cmd/mdsmith` `--build-verify`'s `snapshotOutputs`** — read every
   declared build output fully into memory via `os.ReadFile` to diff two
   verify-pass runs byte-for-byte. Declared outputs are user-defined
   artifacts of unbounded size; this held two full copies of a large output
   per verify pass. Streams each output through sha256 instead. **Behavior
   change, intentional:** the old `bytes.Equal(nil, []byte{})` comparison
   treated a missing output and a present-but-empty one as the same "no
   content" case; the new hash comparison treats them as different, which
   is real non-determinism `--build-verify` exists to catch. Documented on
   `outputHash` and pinned by `TestOutputsEqual_MissingVsEmpty_ReturnsFalse`.
3. **`maxsectionlength` (MDS036) `collectParagraphs`** — ran its own
   `ast.Walk` (duplicating the walk `astutil.CollectSectionParagraphs`
   already performs and memoizes per File for MDS023/MDS024) and called
   `ExtractPlainText` + `CountWords` per paragraph for a value that never
   needed the materialized text. Now delegates to the shared memoized walk
   and uses `mdtext.CountWordsInNode`. **36 → 1 alloc** on a 30-paragraph
   fixture (the single remaining alloc is the presized result slice; the
   walk itself is a cache hit after the first call within a File).
4. **`slidevstructure`/`recipesafety` pure-membership sets** —
   `builtinLayouts`, `knownFMKeys`, and `bodyTemplateReserved` were
   `map[string]bool` read only via truth-tests. Converted to
   `map[string]struct{}` per the guide's "map[K]struct{} for sets" pattern.

A sixth finding — **MDS003/MDS005 each running a redundant full `ast.Walk`
instead of joining the engine's shared `KindScopedChecker` dispatch** — was
the highest-severity hit (both rules are default-enabled and run on every
file with headings), but converting them safely needs a per-file
state-reset hook the `NodeChecker` contract doesn't currently provide (a
rule instance is reused sequentially across files by its worker, and
`CheckNode` has no "start of file" signal). Shipping that conversion
without solving the reset problem risks a state-leak correctness bug
across files, so it's tracked in
[`plan/2608020650_kindscoped-heading-rules.md`](plan/2608020650_kindscoped-heading-rules.md)
instead of included here. Four more rules share the same double-walk shape
and blocker (`blockquotewhitespace`, `codeblockstyle`, `samefileanchor`,
`requiredstructure`) and are noted in that plan as follow-ups.

Each fix follows red/green TDD: a failing allocation/memory-bound test
first, then the minimal fix to make it pass. Three independent xhigh-effort
adversarial review rounds ran against this diff after the initial 5 fixes;
each found real issues (a missing `-race` test skip, an under-exercised
test fixture, a misleading/unnecessary package var, a duplicated AST walk,
a tautological test, an undocumented nil-vs-empty difference, a leftover
`string()` conversion, two inaccurate code comments, and a stale
short-circuit-order regression) — all fixed, commit-by-commit, in the
review-round commits below.

## Test plan

- [x] `go build ./...`
- [x] `go test -race` on every touched package
- [x] `go vet ./...`
- [x] `go tool -modfile=tools/go.mod golangci-lint run` (0 issues)
- [x] `mdsmith check .` (568 files, 0 failures)
- [x] New/updated tests per fix: `markdownflavor` alloc + correctness
      tests (including a lazy-continuation indent edge case) plus a
      `race_on_test.go`/`race_off_test.go` pair, `builddiag` bounded-memory
      test + `outputHash`-typed unit tests including the missing-vs-empty
      distinction, `maxsectionlength` alloc test,
      `slidevstructure`/`recipesafety` zero-byte-set reflection tests

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -259,4 +259,5 @@ footer: |
 | 2607191918 | ✅     | haiku  | [Deduplicate isClaimed between internal/schema and requiredstructure](plan/2607191918_arch-fix-isclaimed-dedup.md)                                      |
 | 2607242010 | 🔲     | sonnet | [MDS072 external-link-check: SSRF and egress hardening](plan/2607242010_mds072-ssrf-network-hardening.md)                                               |
 | 2607242011 | 🔲     | haiku  | [Security hardening batch — 2026-07-24](plan/2607242011_security-hardening-batch-2026-07-24.md)                                                         |
+| 2608020650 | 🔲     | sonnet | [Convert MDS003/MDS005 to KindScopedChecker without a stale-state bug](plan/2608020650_kindscoped-heading-rules.md)                                     |
 <?/catalog?>
--- a/cmd/mdsmith/builddiag.go
+++ b/cmd/mdsmith/builddiag.go
@@ -207,8 +207,16 @@ func verifyTarget(
 }
 
 // outputHash is one declared output's verify-pass fingerprint: ok is
-// false when the file is missing or unreadable (mirroring the old nil
-// []byte), true with hash holding its sha256 otherwise.
+// false when the file is missing or unreadable, true with hash
+// holding its sha256 otherwise. A missing output (ok == false) now
+// compares unequal to a present-but-empty one (ok == true, the sha256
+// of zero bytes) in outputsEqual — a deliberate tightening versus the
+// pre-streaming code, whose bytes.Equal(nil, []byte{}) treated the two
+// as the same "no content" case. A recipe whose first run leaves an
+// output absent and whose second leaves it present-but-empty (or vice
+// versa) is exactly the non-determinism --build-verify exists to
+// catch, so the stricter comparison is intentional (see
+// TestOutputsEqual_MissingVsEmpty_ReturnsFalse).
 type outputHash struct {
 	hash [sha256.Size]byte
 	ok   bool
--- a/cmd/mdsmith/builddiag.go
+++ b/cmd/mdsmith/builddiag.go
@@ -1,8 +1,8 @@
 package main
 
 import (
-	"bytes"
 	"context"
+	"crypto/sha256"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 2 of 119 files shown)


jeduden:

jeduden:
Checked out `claude/kind-darwin-x76fuk` against the current `origin/main` tip (`2ab4b29`, unchanged since this PR branched) — `git merge-tree` reports a clean merge, no conflicting hunks. So there's no real conflict against `main` itself; the queue failure was most likely a transient conflict with another PR that was in the same batch and has since left the queue (no PRs currently carry the `queue` label).

This PR is still a draft, so I'm leaving `queue:failed` as-is rather than re-adding `queue` myself — re-entering the queue is a merge-into-`main` action I'd rather leave to your call once you're ready to take this out of draft. Everything else is green: all 3 rounds of xhigh-effort review plus Copilot's review are addressed and their threads resolved, and `go build/test/vet`, `golangci-lint`, and `mdsmith check .` all pass locally on the current head (`13354cc`).


---

jeduden:
🟢 **Merge Queue** — picked up

**Next:** No action needed — you'll get another comment when CI starts on the batch. [View merge queue run](https://github.com/jeduden/mdsmith/actions/runs/31873601398).

jeduden:

Merged into batch branch [`merge-queue/batch-783-1786781184`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-783-1786781184) alongside #788, #794. [View CI run](https://github.com/jeduden/mdsmith/actions/runs/31873685972).

jeduden:
🔍 **Merge Queue** — bisecting

A larger batch failed CI. Bisection is isolating the culprit: this run tests up to **2 of 3** candidate PRs on [`merge-queue/batch-bisect-783-1786781543`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-bisect-783-1786781543). [View current bisect CI run](https://github.com/jeduden/mdsmith/actions/runs/31873926795).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`a66a32e`](https://github.com/jeduden/mdsmith/commit/a66a32ea0157f210305e5bf1d3bd2cdacc103796). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/31873926795).

PR #785, opened 2026-08-02
perf: audit against high-performance-go.md, fix 5 confirmed hot paths

jeduden:
## Summary

Scheduled performance audit of the codebase against
`docs/development/high-performance-go.md`, following that doc's
"apply best-practice patterns first, then measure" playbook. Five
parallel agents scanned the core lint pipeline, all `internal/rules/*`
packages (split alphabetically), and the supporting/infra packages for
violations of the documented patterns. Findings were ranked by how hot
the call site is and how confidently the fix preserves existing
behavior; the top 5 were fixed, each with a red/green test.

**Three rounds of xhigh-severity adversarial review ran against this
PR, each fixing what the previous round found**, including two cases
where a "fix" was itself measurably wrong — exactly the failure mode
this kind of PR is most at risk of. The corrections are their own
commits so the history shows what was found and how it was fixed,
rather than being squashed away.

Two other candidate findings from the initial scan were investigated
and **deliberately not applied**:

- Adding a raw-byte pre-check before `mdtext.ExtractPlainText` in
  `forbiddentext`/`forbiddenparagraphstarts` (MDS055/MDS056) looked
  promising but is unsafe: markdown markup stripped during plain-text
  extraction (e.g. `wor**d**` → `word`) can join a configured
  substring that never appears contiguously in the raw source, so a
  raw-source pre-check would introduce false negatives.
- A duplicate of the same O(H×P) rescan pattern also exists in
  `occurrence` (MDS060) and `maxsectionlength` (MDS036), using
  different aggregation logic than `astutil.SectionBody`. Left for a
  follow-up rather than risking a rushed fix across three different
  per-range aggregation loops.

## Final fixes (5)

1. **`internal/config`: skip a discarded provenance-selector
   allocation in kind resolution.** `resolveEffectiveKinds` runs on
   every workspace file (it feeds `EffectiveSignature`, the config
   cache key) and, for every matching `kind-assignment` entry, called
   `matchKindAssignmentEntry`, which unconditionally built
   `formatSelector`'s provenance string and threw it away. Split out
   a bool-only `kindAssignmentEntryMatches` for the hot path.
2. **`internal/rules/astutil`: turn `SectionBody`'s O(H×P) per-heading
   rescan into O(H+P).** MDS057/MDS058 call `SectionBody` once per
   heading, each call rescanning the entire paragraphs slice from
   index 0. Heading start lines are non-decreasing in document order,
   so a single forward-only cursor can skip past already-passed
   paragraphs. Added `SectionBodies`, computing every heading's body
   in one sweep, with its per-heading text buffer hoisted and reused
   (`parts[:0]`) rather than re-declared each iteration — the first
   version of this fix actually allocated *more* than the code it
   replaced (10 allocs vs. 6 on a nested-heading fixture); review
   round 3 caught it.
3. **`internal/build`: resolve build-target inputs/outputs once per
   staleness check.** `CheckStaleness` resolved a target's inputs and
   outputs, then called `ComputeActionID`, which re-ran the same
   `resolveInputs`/`resolveOutputs` from scratch. Reuses the
   already-resolved slices via `computeActionIDFromResolved`, matching
   what `RecordBuild` already did.
4. **`internal/config`: skip cloning the discarded side of a
   list-setting merge.** `mergeAny` deep-cloned the *earlier* side of
   a list-typed settings leaf before checking whether the merge mode
   was `Append` or the default `Replace`, discarding the clone on the
   common `Replace` path, then made a second redundant copy of the
   already-cloned `later` side. Check the mode first; only clone
   `earlier` when `Append` needs it; return `later`'s clone directly
   on `Replace` — with an explicit fold back to `nil` for an empty
   later list, since review round 1 caught that the direct-return
   version silently turned a merged `nil` into a non-nil `[]any{}`
   (observable in `mdsmith kinds resolve --json`).
5. **`internal/rules/codeblockstyle`: pre-size `collectBlocksL0`'s
   block slice to the actual code-block count**, not
   `len(BlockSpans)` (which counts every block kind — headings,
   paragraphs, lists, quotes — so presizing to it over-allocates on
   the common case of a mostly-prose file). Counts the real code-block
   spans in a first cheap pass, indexing rather than range-copying the
   32-byte `BlockSpan` struct (review round 3 caught that range-by-value
   made the common case measurably *slower* than doing nothing).

## Reverted from the original top 5

**`internal/discovery`/`internal/globpath`: caching pattern validation
for the `files:` glob walk.** Review round 1 benchmarked it on the
real walk shape and found no significant change (benchstat p=0.912,
above the project's own p<0.05 bar), while the new `MatchRaw` helper
was not actually equivalent to `doublestar.Match` on patterns that
fail `ValidatePattern` but that `Match` still accepts. No live bug at
the one call site, but not worth carrying a correctness trap for a
change with no measured benefit — reverted outright and replaced by
fix 5 above.

## What each review round found and fixed

- **Round 1:** nil-vs-`[]` correctness regression in fix 4's replace
  path (fixed); no measured benefit + a semantic gap in the
  discovery/globpath change (reverted); missing precondition docs on
  `SectionBodies` (added).
- **Round 2:** fix 5's first draft (presize to `len(BlockSpans)`) was
  a measured regression on the common case — ~20x over-allocation,
  ~14x slower on a prose-only fixture (fixed with a count-first pass);
  the `SectionBodies` precondition doc named only paragraph ordering,
  missed that headings must also be sorted (added); an undocumented
  peak-live-memory trade-off in the batch API (documented).
- **Round 3:** fix 5's count-first pass itself range-copied the
  32-byte `BlockSpan` struct, making the *fix for the regression*
  ~1.85x slower than doing nothing on the common case (fixed by
  indexing instead); fix 2's hoisted `parts` buffer wasn't actually
  hoisted, so it allocated *more* than the O(H×P) code it replaced
  (fixed); a doc comment claiming unchanged allocation volume was
  measurably false (corrected); added a consistency test between two
  hand-maintained type switches (`isAnySliceType`/`toAnySlice`) that
  had nothing enforcing they stay in sync.

Every fix above and every correction was independently verified with
its own differential/fuzz test, allocation-budget test, or benchmark —
not just asserted. Full details are in the individual commit messages.

## Test plan

- [x] Each fix has a dedicated red→green test (`testing.AllocsPerRun`
      budget test, or an equivalence/call-count test where allocation
      counting doesn't capture the win) written before the fix, shown
      failing, then passing after.
- [x] `go build ./...`
- [x] `go test ./...` (all 161 packages pass)
- [x] `go vet ./...`
- [x] `go tool -modfile=tools/go.mod golangci-lint run ./...` (0 issues,
      full repo)
- [x] `go run ./cmd/mdsmith check .` (567 files checked, 0 failures)
- [x] `internal/integration` corpus/allocation-budget suites pass
      unchanged
- [x] Three xhigh-severity adversarial review rounds, each addressing
      every confirmed finding from the previous round

Ref: `docs/development/high-performance-go.md`

### DIFF
--- a/internal/build/staleness.go
+++ b/internal/build/staleness.go
@@ -23,6 +23,13 @@ var globCapFn = buildpathutil.CheckGlobMatchCap
 // hashFileFn is the hashFile implementation; tests may replace it.
 var hashFileFn = hashFile
 
+// resolveInputsFn and resolveOutputsFn are the resolveInputs/resolveOutputs
+// implementations; tests may replace them to count calls.
+var (
+	resolveInputsFn  = resolveInputs
+	resolveOutputsFn = resolveOutputs
+)
+
 // Verdict is the staleness result for one target.
 type Verdict int
 
@@ -226,10 +233,10 @@ func computeActionIDFromResolved(in StalenessInput, inputs, outputs []string) (s
 // --build-no-cache) to catch configuration errors cheaply, without the I/O
 // cost of a full staleness check.
 func ValidateInputs(in StalenessInput) error {
-	if _, err := resolveInputs(in); err != nil {
+	if _, err := resolveInputsFn(in); err != nil {
 		return err
 	}
-	_, err := resolveOutputs(in)
+	_, err := resolveOutputsFn(in)
 	return err
 }
 
@@ -239,11 +246,11 @@ func ValidateInputs(in StalenessInput) error {
 // with an outer 8-byte big-endian length; nested keys, values, and paths
 // are themselves framed. Returns "sha256-<64 lowercase hex>".
 func ComputeActionID(in StalenessInput) (string, error) {
-	inputs, err := resolveInputs(in)
+	inputs, err := resolveInputsFn(in)
 	if err != nil {
 		return "", err
 	}
-	outputs, err := resolveOutputs(in)
+	outputs, err := resolveOutputsFn(in)
 	if err != nil {
 		return "", err
 	}
@@ -272,11 +279,11 @@ type ActionExplanation struct {
 
 // Explain resolves a target's ActionID inputs and returns the breakdown.
 func Explain(in StalenessInput) (ActionExplanation, error) {
-	inputs, err := resolveInputs(in)
+	inputs, err := resolveInputsFn(in)
 	if err != nil {
 		return ActionExplanation{}, err
 	}
-	outputs, err := resolveOutputs(in)
+	outputs, err := resolveOutputsFn(in)
 	if err != nil {
 		return ActionExplanation{}, err
 	}
@@ -317,11 +324,11 @@ func hashFile(abs string) (string, error) {
 // output → stale; compute ActionID; cache miss or different ActionID →
 // stale; any output content hash mismatch → stale; otherwise fresh.
 func CheckStaleness(in StalenessInput, cache *Cache) (StalenessResult, error) {
-	inputs, err := resolveInputs(in)
+	inputs, err := resolveInputsFn(in)
 	if err != nil {
 		return StalenessResult{}, err
 	}
-	outputs, err := resolveOutputs(in)
+	outputs, err := resolveOutputsFn(in)
 	if err != nil {
 		return StalenessResult{}, err
 	}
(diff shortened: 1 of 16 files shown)


jeduden:
🔍 **Merge Queue** — bisecting

jeduden:
⏳ **Merge Queue** — requeued

jeduden:

Merged into batch branch [`merge-queue/batch-775-1786780171`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-775-1786780171) alongside #775, #780, #782, #786. [View CI run](https://github.com/jeduden/mdsmith/actions/runs/31872985040).

jeduden:
✅ **Merge Queue** — merged

PR #786, opened 2026-08-03
perf: fix top 5 high-performance-go.md violations

jeduden:
## Summary

An audit against [`docs/development/high-performance-go.md`](https://github.com/jeduden/mdsmith/blob/main/docs/development/high-performance-go.md) — four parallel scans across allocation/regex patterns, string/bytes patterns, struct layout, and concurrency/misc anti-patterns. The codebase is already unusually disciplined about these guidelines (most hot paths cite the doc directly, and `internal/integration/alloc_budget_test.go` gates the ≤10-alloc rule budget), so genuine violations were sparse. This PR fixes the top 5 found, ranked by how hot the code path is and how safe the fix is to land.

## Fixes

1. **Struct layout: `anchorEntry` / `EmphasisParagraph`** — both declared a scalar field before a pointer-containing one, so the GC's ptrdata scan covered the whole struct instead of stopping after the last pointer field. `anchorEntry` backs MDS027 (default-on); `EmphasisParagraph` backs MDS018 (default-on). Reordered to match the sibling `runCacheEntry` type that already gets this right, and pinned with `structlayout.AssertPointerFieldsFirst` regression tests. ("Struct layout")

2. **Hand-rolled newline scan → `bytes.IndexByte`** — `collectDefinitions` (MDS053, default-on) and `collectReferenceDefinitions` (MDS043) each hand-rolled a byte-by-byte loop over the whole file `Source` to find each line's trailing newline. `bytes.IndexByte` is SIMD-accelerated and the compiler doesn't vectorize a hand-rolled scan loop; `internal/lint/linkrefscan.go`'s `buildLineSegments` already documents and uses the `IndexByte` form for the identical whole-document scan — these two rules had independently reimplemented the line splitter and regressed to the slower loop. Added a benchmark per package over a 500-paragraph fixture. ("Strings and bytes")

3. **`fmt.Sprintf("%q", x)` → `strconv.Quote(x)`** — `fmt.Sprintf` goes through reflection; for a plain string quote, `strconv.Quote` produces byte-identical output roughly 3x faster with no reflection. Swapped every such call site in the diagnostic-construction paths of `catalog`, `concisenessscoring`, `paragraphstructure`, `requiredstructure`, and the shared `schema` package. The existing test suite already asserts the exact quoted message text these sites produce, so it's the correctness gate for the swap. ("Allocations")

4. **Reflect-based rule cloning cache for `RunSource` (the LSP hot path)** — `runFiles`' per-worker `confCache` already spares a corpus-wide `mdsmith check` from re-cloning a `Configurable` rule (`rule.CloneRule`, a `reflect.New` + `ApplySettings` pair) more than once per distinct effective config, because one worker walks many files sequentially. `RunSource` / `RunSourceWithVersion` had no equivalent: the LSP's per-keystroke `CheckVersion` call re-paid that reflect-based clone on *every* debounced edit for every `Configurable` rule with settings, even though the session's effective config never changes between edits. Added `SourceConfigCache`, an opt-in `Runner` field (nil by default — every non-LSP caller is unaffected) keyed by `config.EffectiveSignature`, the same signature `runFiles` already uses for its own cache. `pkg/mdsmith.Session` installs one alongside its existing `RunCache` and `ParseCache`. ("Skip work you don't need" / reflect in hot paths)

## Motivation

`docs/development/high-performance-go.md` states the process explicitly: apply the documented best-practice patterns first, since they cost nothing and are known wins across the Go core. Each fix here closes a concrete, cited gap between that document's guidance and the current code, on paths that run once per file (or once per LSP keystroke) across the whole workspace.

## Test plan

- [x] Red/green TDD for each fix: a failing `structlayout` test / build error / benchmark came first, then the fix, then green.
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` (full suite, including the alloc-budget gate in `internal/integration`)
- [x] `go tool -modfile=tools/go.mod golangci-lint run` on every touched package — 0 issues
- [x] `go run ./cmd/mdsmith check .` — 567 files checked, 0 failures

---
_Generated by [Claude Code](https://claude.ai/code/session_01HPgd5MgHTQAGnWbjgGy3c4)_

### DIFF
--- a/internal/engine/runner.go
+++ b/internal/engine/runner.go
@@ -153,6 +153,14 @@ type Runner struct {
 	// hit, and that RunCache's Invalidate seam would be the one
 	// driving correctness — easy to get wrong, so don't mix.
 	ParseCache *lint.ParseCache
+	// SourceConfigCache memoizes checker.ConfigureEnabledRules for
+	// RunSource / RunSourceWithVersion, keyed by the effective-config
+	// signature (see SourceConfigCache's own doc comment). nil (the
+	// default) leaves RunSource configuring fresh on every call, exactly
+	// as before this field existed. A long-lived caller — the LSP session
+	// — installs one so a buffer re-linted at an unchanged effective
+	// config skips re-cloning every Configurable rule's settings.
+	SourceConfigCache *SourceConfigCache
 }
 
 // fileOutcome is one file's contribution to a run. Workers fill a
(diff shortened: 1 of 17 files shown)


jeduden:
🔍 **Merge Queue** — bisecting

jeduden:
⏳ **Merge Queue** — requeued

jeduden:

Merged into batch branch [`merge-queue/batch-775-1786780171`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-775-1786780171) alongside #775, #780, #782, #785. [View CI run](https://github.com/jeduden/mdsmith/actions/runs/31872985040).

jeduden:
✅ **Merge Queue** — merged

PR #790, opened 2026-08-08
perf: audit against high-performance-go.md, fix top 5 hot-path violations

jeduden:
## Summary

Audited the codebase against `docs/development/high-performance-go.md` using five parallel research agents, each covering a different subsystem (`internal/rules`, `internal/lint`/`engine`/`discovery`, `internal/output`/`export`/`metrics`/`linkgraph`, `cmd/mdsmith`/`lsp`/`build`, `pkg/markdown`/`schema`). Most of the codebase is already heavily optimized from prior audit passes (buffer pools, byte-native scans, `RunCache` memoization, struct field ordering) and golangci-lint already enforces `perfsprint`/`prealloc`/gocritic's performance checks in CI, so this pass targeted the subtler violations lint doesn't catch. Ranked the ~25 candidate findings by hot-path frequency and blast radius and fixed the top 5, each with red/green TDD:

1. **`reflect` in hot Check/Fix paths** — `sort.Slice`/`sort.SliceStable` drive `reflect.Swapper` internally (the doc's own "reflect in hot paths" anti-pattern). Replaced with `slices.Sort(Stable)Func` across MDS062 (link-validity), the `astutil` section-heading helper shared by MDS057/MDS058, MDS019 (catalog sort key), and MDS034 (markdownflavor findings/edits + the shared `flavor.Detect` sort). Measured: each sort now lands at 0 allocs beyond one-time slice setup, versus ~3 `reflect.Swapper` allocs per call before.
2. **`ExtractWikiLinks` (MDS027) ungated collection** — every call built the code-block, PI-block, and code-span line-set maps and ran the wikilink regex over the whole source, even though wikilinks are an Obsidian-specific extension almost no file uses. Added a `bytes.Contains(f.Source, "[[")` pre-check per the doc's "gate expensive analyzers behind a cheap pre-check" pattern (the same one MDS024/MDS043 already use). Measured: 0 allocs on a warm file with no wikilinks, versus building all three line-set maps before.
3. **Bare-URL regex with no pre-check gate** — `BareURLFindingsInTree` ran the multi-alternative bare-URL regex against every Text node unconditionally whenever the flavor accepts autolinks (the GFM/goldmark/Pandoc default). Added a `bytes.Contains(body, "://")` gate. Measured: 0 allocs on scheme-free prose, versus running the regex engine on every Text node before.
4. **Discarded provenance string on the per-file kind-resolution path** — `resolveEffectiveKinds` (backing `EffectiveSignature`, called unconditionally once per file including on a config-signature cache hit) called `matchKindAssignmentEntry`, which always built a `strings.Join`-based selector string, then threw it away with `_`. Split into a match-only `kindAssignmentEntryMatches` for the hot path and kept the selector-building version for `provenance.go`'s cold explain path, its only real consumer.
5. **`doublestar.Match`'s per-call pattern revalidation** — the workspace file-discovery walk's `matchesAny` called `doublestar.Match`, which revalidates its pattern on every call, even though `Discover` already validates every pattern once via `validatePatterns` before the walk starts. `internal/globpath` fixed the identical issue in the config-glob path with a documented ~4%-of-check-CPU measurement; `discovery.go` imports doublestar directly and was never migrated. Switched to `MatchUnvalidated`. Measured ~270ns/op → ~155ns/op on a representative no-match case; confirmed the new regression gate (`TestMatchesAny_NoRevalidationBudget`) is red on the unfixed code before landing.

Each fix ships a dedicated allocation or timing regression test (`alloc_test.go` / `*_alloc_test.go` / `matches_bench_gate_test.go`) following the project's existing `testing.AllocsPerRun`/`testing.Benchmark`-in-a-`Test` conventions, so a future regression fails `go test ./...` on its own.

## Test plan

- [x] `go build ./...`
- [x] `go test ./...` (full suite, including `internal/integration`'s per-rule alloc budget gate)
- [x] `go test -race ./...` on every touched package
- [x] `go vet ./...`
- [x] `mdsmith check .` — 567 files checked, 0 failures
- [x] Manually confirmed each new perf gate is red on the pre-fix code (verified for the sort removal, the wikilink/bare-URL gates, and the discovery benchmark) and green after the fix

Claude-Session: https://claude.ai/code/session_01BxqcjhN1RmEkXdDo6WU5SR

---
_Generated by [Claude Code](https://claude.ai/code/session_01BxqcjhN1RmEkXdDo6WU5SR)_


jeduden:
🔍 **Merge Queue** — bisecting

jeduden:
⏳ **Merge Queue** — requeued

jeduden:
Superseded by #800, which consolidates this PR with #776, #779 and #793 and resolves every conflict by hand.

Carried forward from this PR:

- `astutil.sortSectionHeadings`, `flavor.sortFindingsByStart`, and the other `slices.SortStableFunc` conversions that drop `reflect.Swapper` from hot Check/Fix paths
- `linkvalidity`, `markdownflavor/fix.go`, `linkgraph` and `flavor/detect.go` changes, plus the alloc tests (unioned with `main`'s where both sides added one)
- `config/kindmatch.go`'s selector-free match path

Combined rather than picked, in `catalog/rule.go`: this PR moves the sort to `slices.SortStableFunc` (no reflect) but re-parses the sort key inside the comparator, which is O(n log n) parses. `main` pre-computes `numInt` once per entry but still sorts through reflect. #800 takes both — `slices.SortStableFunc` reading the pre-computed `numInt` — so `sortSortables` drops its now-unused `key` parameter and `sortentries_alloc_test.go` is updated to match.

Not carried forward:

- `discovery.matchesAny`: this PR has the unconditional `MatchUnvalidated`. #779 showed that diverges from `Match` on brace alternatives containing a syntax error, so #800 takes #779's brace-guarded version.
- `internal/discovery/matches_bench_test.go`: its `BenchmarkMatchesAny` duplicates `main`'s, which is more faithful because it builds `w.patterns` through `validatePatterns`.

PR #792, opened 2026-08-09
perf: stop reallocating a disabled logger every file (Runner + Fixer)

jeduden:
## Summary

Scheduled audit against `docs/development/high-performance-go.md`, per the recurring "find and fix top violations" routine (following PRs #762, #764, #765, #767).

**Methodology:** five parallel static-pattern agents scanned the codebase for the doc's named anti-patterns (regexes compiled per-call, `string(b)` conversions in hot paths, struct-layout/pointer-slice issues, missing memoization, loop/concurrency anti-patterns) — all five came back clean. The codebase has already been through four prior rounds of this exact audit and is well-optimized: allocation budgets are enforced by CI (`internal/integration/alloc_budget_test.go`), `golangci-lint` already runs `perfsprint`/`prealloc`/gocritic's performance group, and most hot paths carry explicit comments citing the plan that already fixed them.

Given the static sweep found nothing, I profiled the `BenchmarkCheckCorpusLarge` benchmark directly (CPU + alloc profiles over mdsmith's own 600+ file doc corpus) to find anything static pattern-matching would miss. That surfaced one confirmed, safe fix, and a code-review pass then found the identical bug in a sibling package:

- **`internal/engine`: `Runner.log()` allocated a fresh empty `*vlog.Logger` on every per-file call** when no `Logger` was configured (the default, non-`-v` case). `lintFile` calls `log()` once per file, so this was one wasted allocation per workspace file for a value nothing ever mutates. Fixed by sharing one disabled-logger singleton — safe because every disabled `Logger` is behaviorally identical (`Printf` no-ops without touching `W` or `mu`) and it's never mutated through the returned pointer.
- **`internal/fix`: `Fixer.log()` had the exact same bug**, found during the code-review pass on the first fix — the `fixOnce` loop calls `log()` once per file with the identical fallback-allocation pattern. Fixed the same way.

**One candidate was investigated and rejected after measurement.** Three default-on rules (`notrailingspaces`, `nohardtabs`, `nomultipleblanks`) do a `map[int]struct{}` lookup per line to skip code-block lines — CPU profiling attributed ~75-85% of each rule's `Check` cost to that lookup, which looked like a textbook match for the doc's "sorted slice + binary search beats a map for n < ~100" guidance. I implemented the swap (`lint.SortedCodeBlockLines` + binary search) and benchmarked it with `benchstat` before committing, per the doc's own "decide with benchstat, not eyeballs" rule — it was a **regression** (+9.6% CPU, p=0.000, n=10; +4.8% allocs), because building the sorted slice fresh per `File` outweighs the saved per-line hash lookups at this rule set's typical code-block density. Reverted rather than shipped; documenting here so a future audit doesn't retread it without re-measuring at a different corpus profile.

## Changes

- `internal/engine/runner_log.go` / `internal/engine/runner_log_test.go`: share a `disabledLogger` singleton instead of allocating `&vlog.Logger{}` per call; test pins the shared-instance behavior and a deterministic 0-allocs/op budget.
- `internal/fix/fix.go` / `internal/fix/fix_log_test.go`: identical fix and test shape for `Fixer.log()`.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go tool -modfile=tools/go.mod golangci-lint run` (0 issues)
- [x] Both new tests confirmed red against pre-fix code, green after (TDD)
- [x] Code review run at xhigh severity (2 passes so far, addressing findings between runs)

### DIFF
--- a/internal/engine/runner_log.go
+++ b/internal/engine/runner_log.go
@@ -9,22 +9,14 @@ import (
 	"github.com/jeduden/mdsmith/internal/rule"
 )
 
-// disabledLogger is the shared fallback log() returns when no Logger
-// is configured. lintFile calls log() once per file, so allocating a
-// fresh *vlog.Logger there would cost one alloc per workspace file for
-// a value nothing ever mutates; every disabled Logger is
-// behaviorally identical (Printf no-ops without touching W or mu), so
-// one shared instance is safe to hand out to every caller, including
-// concurrently from the parallel lint pipeline.
-var disabledLogger = &vlog.Logger{}
-
-// log returns the runner's logger. If no logger is set, it returns a
-// disabled logger so callers don't need nil checks.
+// log returns the runner's logger. If no logger is set, it returns
+// vlog.Disabled — a shared instance, not a fresh allocation, since
+// lintFile calls log() once per file (see vlog.Disabled's doc).
 func (r *Runner) log() *vlog.Logger {
 	if r.Logger != nil {
 		return r.Logger
 	}
-	return disabledLogger
+	return vlog.Disabled
 }
 
 // logRules logs each enabled rule in the effective config from the provided slice.
--- a/internal/engine/runner_log.go
+++ b/internal/engine/runner_log.go
@@ -9,13 +9,22 @@ import (
 	"github.com/jeduden/mdsmith/internal/rule"
 )
 
+// disabledLogger is the shared fallback log() returns when no Logger
+// is configured. lintFile calls log() once per file, so allocating a
+// fresh *vlog.Logger there would cost one alloc per workspace file for
+// a value nothing ever mutates; every disabled Logger is
+// behaviorally identical (Printf no-ops without touching W or mu), so
+// one shared instance is safe to hand out to every caller, including
+// concurrently from the parallel lint pipeline.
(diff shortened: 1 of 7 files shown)


jeduden:
🔍 **Merge Queue** — bisecting

jeduden:
⏳ **Merge Queue** — requeued

jeduden:
Closing as superseded by #770.

Both hot-path fixes this PR makes have already landed on `main`, implemented differently:

- `internal/engine`: `disabledLogger` package var (replaces the per-file `&vlog.Logger{}` in `(*Runner).log`)
- `internal/fix`: `disabledFixerLogger` package var (same fix in `(*Fixer).log`)

After discarding what is already on `main`, the residual content is:

- `internal/log/logger.go` — the exported `vlog.Disabled` var, a consolidation of the two package-level vars. A refactor, not a perf gain.
- `internal/log/logger_test.go`, `internal/fix/fix_log_test.go` — tests for the above.
- `internal/lsp/server.go` — one line. That call site is inside `lsp.New()`, a constructor invoked once per server start, so it is not a hot path.

If the `vlog.Disabled` consolidation is still wanted as a cleanup, it is worth a small standalone PR against current `main` rather than carrying this branch forward.

PR #800, opened 2026-08-15
perf: consolidate #776, #779, #790, #793 with manual conflict resolution

jeduden:
Consolidates the last four open `perf:` PRs from the recurring
high-performance-go.md audit into one branch, with every conflict
resolved by hand.

Supersedes #776, #779, #790, #793 — close those in favour of this.
(#792 was closed separately as fully superseded by #770.)

## Why one PR

All four were opened against a `main` that had not moved in three
weeks, so each audit rediscovered overlapping hot spots. Landing them
individually meant four rebase cycles over the same files
(`astutil.go`, `catalog/rule.go`, `discovery.go`,
`flavor/detect.go`), with each merge invalidating the next.

## Conflict resolutions

Every conflict was read individually. Three kinds came up:

**Same fix, better implementation already on `main`** — took `main`:

- `markdownflavor/rule.go`: `main`'s `f.LineOfOffset` + `bytes.Buffer`
  over #776's `flavor.LineCol` + pre-sized `[]string`+join.
- `config/kindmatch.go`, `linkgraph/wikilinks.go`,
  `flavor/detect.go`: comment-only differences around identical code.

**Both sides contribute something the other misses** — combined:

- `catalog/rule.go`: #790 drops `reflect.Swapper` by moving to
  `slices.SortStableFunc`, but re-parses the sort key inside the
  comparator (O(n log n) parses). `main` pre-computes `numInt` once
  per entry but still sorts through reflect. The merge takes both:
  `slices.SortStableFunc` reading the pre-computed `numInt`.
  `sortSortables` loses its now-unused `key` parameter.
- `astutil.go`: union of #776's `CollectHeadingNodes` and #790's
  `sortSectionHeadings` — unrelated additions git could not align. The
  `sort` import stays for `SectionBody`'s `sort.Search`, which is not
  the reflect-based `sort.Slice` #790 targeted.

**A correctness fix on top of the optimization** — took the PR:

- `discovery.go` `matchesAny`: `main` calls
  `doublestar.MatchUnvalidated` unconditionally. That diverges from
  `Match` on brace alternatives whose non-final alternative contains a
  syntax error. #779 keeps the fast path for brace-free patterns (the
  common case) and falls back to validated `Match` when the pattern
  contains `{`. **This fixes a latent bug on `main`, not just a
  conflict.**

## Dropped as redundant

- #779's `BenchmarkWalker_matchesAny` and #790's
  `matches_bench_test.go` both duplicate `main`'s
  `BenchmarkMatchesAny`, which is more faithful — it builds
  `w.patterns` through `validatePatterns`, as `Discover` does.
- #779's `TestBaseNode_PointerFieldsBeforeScalars`, already on `main`
  in `basenode_structlayout_test.go`.

Add/add conflicts in `kindmatch_alloc_test.go`,
`linkvalidity/alloc_test.go` and `markdownflavor/alloc_test.go` are
unions — each side adds differently-named tests. #776's alloc budgets
were re-run against `main`'s implementation and still hold.

## One correction worth flagging

Combining #776's memoized `CollectHeadingNodes` with `main`'s
`HeadingTextCached` in MDS005 pushed it to **12 allocs/op** against
CLAUDE.md's documented ceiling of 10: each helper installs its own
per-File memo, and a rule measured in isolation pays for both without
amortising either across sibling rules. Each alone fits. The final
commit keeps #776's shared heading walk (the point of that PR — MDS003
uses it too) with the direct `HeadingText` call, as #776 shipped it.

Caught by `TestPerRuleAllocBudget`.

## Verification

Against this branch merged onto `main`:

- `go build ./...`, `go vet ./...`, `go vet -tags race ./...` — clean
- `golangci-lint run` — 0 issues
- `gofmt -l` — clean
- Full test suite — clean
- `internal/integration` failures and `mdsmith check` output are
  **byte-identical to clean `origin/main`** in this environment (three
  packages fail on bare `main` in a detached worktree because they
  shell out to `go` and misresolve the module root)


jeduden:
🟢 **Merge Queue** — picked up

**Next:** No action needed — you'll get another comment when CI starts on the batch. [View merge queue run](https://github.com/jeduden/mdsmith/actions/runs/31897637949).

jeduden:

Merged into batch branch [`merge-queue/batch-800-1786813970`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-800-1786813970). [View CI run](https://github.com/jeduden/mdsmith/actions/runs/31897682074).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`1c8b415`](https://github.com/jeduden/mdsmith/commit/1c8b4151e01bec85cea54e7fba8814fe56525b9e). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/31897682074).

PR #811, opened 2026-08-17
perf: fix top 5 high-performance-go.md violations

jeduden:
## Summary

A multi-agent audit of the codebase against
`docs/development/high-performance-go.md` found the project already
highly compliant with its own patterns (pre-sized slices, pooled
buffers, package-scope regexes, byte-gated cheap pre-checks). The
handful of genuine violations found were ranked by severity/impact —
weighted toward default-enabled rules and per-invocation hot paths —
and the top 5 are fixed here with red/green TDD and measured
before/after benchmarks where the win was CPU-bound rather than
allocation-bound. Three rounds of xhigh-severity code review then
surfaced and fixed two more instances of the same anti-pattern plus
one real correctness bug introduced by the new cache (see below).

1. **MDS019 (catalog) — redundant AST walk, no cheap gate.**
   `Check` unconditionally ran `gensection.FindMarkerPairs`'
   top-level AST walk three times per file (once via the engine,
   once each in `checkCaseMismatches` and `checkInjection`) even on
   files with no `<?catalog?>` directive at all — the common case,
   since MDS019 is default-enabled. Added a `bytes.Contains`
   pre-check mirroring `include`'s existing gate, and share one
   `FindMarkerPairs` result between the two hint passes. An
   isolated-timer benchmark shows Check's own cost drops ~75%
   (6.4us → 1.6us, p=0.000) on a no-directive file.

2. **`internal/engine` — reflect-based sort on every run's full
   diagnostic set.** `sortDiagnostics` (the final merge sort every
   `Runner.Run`/`RunSource` call performs) used `sort.SliceStable`,
   which drives `reflect.Swapper` internally — the one place in the
   codebase this exact fix (`slices.SortStableFunc`) had not yet been
   applied, unlike five other packages that sort `lint.Diagnostic`
   the same way. A benchmark sorting 2000 diagnostics shows -14.46%
   CPU and -66.67% allocations (p=0.000).

3. **MDS025 (table-format) — same reflect-sort anti-pattern.** Both
   `Check`'s final merge sort and `structureDiagnostics`' own sort
   used `sort.SliceStable`; MDS025 is default-enabled and both run on
   any file containing a `|` character. Switched to
   `slices.SortStableFunc`.

4. **Remaining reflect-based sorts.** Six packages
   (`ambiguousemphasis`, `noinlinehtml`, `propernames`,
   `nounusedlinkdefinitions`, and — found in review —
   `noreferencestyle`, `nospaceincodespans`) still sorted
   diagnostic-adjacent slices via `sort.Slice`/`sort.SliceStable`.
   Completed the fix everywhere it was still missing.

5. **MDS020 (required-structure) — schema re-read + re-parsed per
   host file.** `dispatchSingleFileSchema` read the schema file from
   disk and re-unmarshalled its YAML front matter (to peek at
   `extends:`) on every single host file's `Check` call. For a
   workspace-wide kind, N host files paid for N identical reads of
   the same schema. Added `RunCache.RawSchemaFile`, a new cache slot
   mirroring the existing `FrontMatter`/`ParsedSchema` slots (same
   load-once-per-absPath semantics, same `Invalidate` wiring), and
   routed the read + extends-peek through it. Verified with a
   counting-`fs.FS` test: the schema is now read once across three
   host files sharing a `RunCache`, and a fresh read is forced again
   after `Invalidate`.

Each fix references the specific guideline pattern it applies from
`docs/development/high-performance-go.md` ("gate expensive analyzers
behind a cheap pre-check", "reflect in hot paths", "memoize
per-input computations").

### Review fixes

Three rounds of xhigh-severity review found and fixed, in order:

- A missing coverage case for the new Line-tiebreak branch in
  `ambiguousemphasis`'s sort comparator (Codecov patch-coverage gate).
- Two `sort.Slice` cut-sorters (`noreferencestyle`,
  `nospaceincodespans`) that follow the exact pattern item 4 was
  meant to close out, but were missed in the original sweep.
- A real bug in the new MDS020 cache: `cachedRawSchema` keyed its
  slot only by absolute schema path, but cached an error string
  already formatted with the *first* caller's `schemaPath` spelling.
  Two kinds referencing the same missing schema via differently
  spelled but path-equivalent strings (`docs/proto.md` vs
  `./docs/proto.md`) would then share a cache slot, and the second
  kind's diagnostic would silently report the first kind's spelling.
  Fixed by caching only the raw (already-normalized) read error and
  formatting the caller-facing message at each call site from that
  caller's own argument. Reproduced with a regression test before
  fixing.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` (full suite, no regressions)
- [x] `go tool -modfile=tools/go.mod golangci-lint run` on every
      touched package (0 issues)
- [x] `internal/integration` per-rule allocation budget gate
      (`TestPerRuleAllocBudget`) — all rules, including MDS019,
      MDS020, MDS025
- [x] `mdsmith check .` — self-lint passes clean
- [x] New unit/behavioral tests for each fix (red before, green
      after), plus targeted benchmarks with `benchstat` deltas for
      the two CPU-bound fixes (catalog gate, engine sort)
- [x] Three rounds of xhigh-severity code review, all flagged issues
      fixed and re-verified

### DIFF
--- a/internal/engine/bench_sort_test.go
+++ b/internal/engine/bench_sort_test.go
@@ -1,0 +1,40 @@
+package engine
+
+import (
+	"strconv"
+	"testing"
+
+	"github.com/jeduden/mdsmith/internal/lint"
+)
+
+// BenchmarkSortDiagnostics pins the CPU cost of sorting a
+// representative-size diagnostic set from a large workspace run.
+// sortDiagnostics runs once per Runner.Run/RunSource call over the
+// full result set (internal/engine/runner.go's two call sites) — the
+// one place in the codebase where the reflect-vs-slices.SortStableFunc
+// fix from docs/development/high-performance-go.md's "reflect in hot
+// paths" anti-pattern had not yet been applied, unlike the equivalent
+// per-rule sortDiagnostics helpers in linkvalidity, catalog, astutil,
+// duplicatedcontent, and markdownflavor.
+func BenchmarkSortDiagnostics(b *testing.B) {
+	const n = 2000
+	base := make([]lint.Diagnostic, n)
+	for i := range base {
+		// Descending input order forces real comparison work instead
+		// of a sort implementation detecting an already-sorted run.
+		base[i] = lint.Diagnostic{
+			File:    "file" + strconv.Itoa(n-i) + ".md",
+			Line:    n - i,
+			Column:  1,
+			RuleID:  "MDS001",
+			Message: "message",
+		}
+	}
+
+	diags := make([]lint.Diagnostic, n)
+	b.ReportAllocs()
+	for i := 0; i < b.N; i++ {
+		copy(diags, base)
+		sortDiagnostics(diags)
+	}
+}
--- a/internal/engine/runner_log.go
+++ b/internal/engine/runner_log.go
@@ -1,7 +1,8 @@
 package engine
 
 import (
-	"sort"
+	"cmp"
+	"slices"
 
 	"github.com/jeduden/mdsmith/internal/config"
 	"github.com/jeduden/mdsmith/internal/lint"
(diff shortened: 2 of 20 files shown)


jeduden:
🟢 **Merge Queue** — picked up

**Next:** No action needed — you'll get another comment when CI starts on the batch. [View merge queue run](https://github.com/jeduden/mdsmith/actions/runs/33241035948).

jeduden:

Merged into batch branch [`merge-queue/batch-811-1787988713`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-811-1787988713) alongside #812, #815. [View CI run](https://github.com/jeduden/mdsmith/actions/runs/33241089124).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`5970ff2`](https://github.com/jeduden/mdsmith/commit/5970ff2aea8877bcc80e2e811b657ea041d00f54). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/33241089124).

PR #813, opened 2026-08-20
perf: cut check allocations 21% and CPU 6% across five hot-path fixes

jeduden:
## Overview

An agent-team audit of the Go core against
[docs/development/high-performance-go.md](https://github.com/jeduden/mdsmith/blob/main/docs/development/high-performance-go.md)
ranked every violation it could verify by measured impact. The top five
are fixed here, each with red/green TDD and a before/after measurement.

Findings came from four parallel agents working different angles —
allocation patterns, redundant work, data structures, and pure
profiling — then cross-checked. Every number below is measured: CPU
from merged `MDSMITH_CPUPROFILE` runs, allocations from
`BenchmarkCheckCorpusLarge` (600 files), end-to-end timings from an
interleaved A/B of paired `mdsmith check .` runs on this repository.

Work already covered by the open #811 was excluded, so these five do
not overlap it.

**Net effect** — re-measured against the current base (`348d95a`) after
merging main in:

| Metric | main | this branch | Delta |
| --- | --- | --- | --- |
| allocs/op (600-file corpus) | 127,000 | 100,000 | **-21.3%** |
| B/op (600-file corpus) | 74.4 M | 69.1 M | **-7.1%** |
| `check .` user+sys, median of 25 | 0.700 s | 0.650 s | **-7.1%** |

The branch was faster in 25 of 25 paired runs, and `mdsmith check .`
output is byte-identical to `main`.

### 1. `linkgraph` built every link's visible text and nobody read it

`ExtractLinks` / `ExtractImages` / `ExtractRefLinkTargets` called
`mdtext.ExtractPlainText` per link node, costing a `bytes.Buffer` plus
a string copy each time. Nothing on the `check` path reads the label —
MDS027, MDS068 and the index builder use only `Line`, `Column` and
`Target`. The sole reader is `mdsmith list backlinks`.

The eager `Link.Text` field became a `Link.Text(source)` method backed
by the originating AST node. `RefLink` had the same waste with *no*
reader at all, so its text field is gone outright.

*Guideline:* #skip-work-you-dont-need. **-6.1% allocs/op.**

### 2. Link destinations went through `string()` + `net/url.Parse`

Every link and image paid a string copy and a full `url.URL` allocation
to recover a path and a fragment — or just to be discarded as external.

New `ParseTargetBytes` keeps `ParseTarget` as the reference
implementation and takes two shortcuts that cannot change an answer: an
external destination is rejected straight from the bytes with no string
copy, and an ordinary local one is split on `#` into substrings of the
single string it allocates. Anything unusual — percent-escape, query,
control character, colon — falls through to `ParseTarget` unchanged.

Equivalence is pinned by an exhaustive differential test over all 271k
strings up to length 5 drawn from the characters that drive scheme
detection, escaping, query stripping and control rejection. That
enumeration caught two real mismatches while the fix was being written
(`net/url` rejects a colon in the first path segment), which is why
colons route to the full parse.

*Guideline:* "Stay in `[]byte`". **-6.6% allocs/op.**

### 3. MDS056 rescanned every paragraph once per configured needle

`forbidden-text` ran one end-to-end `strings.Contains` per needle on
every paragraph. Under the `no-llm-tells` convention (~70 needles,
pinned by this repo) that is ~70 full scans of every paragraph in the
workspace; a merged CPU profile attributed **9.4% of all `check` CPU**
to the rule, almost entirely inside `strings.Contains`.

The needle list now compiles into a fully-folded Aho-Corasick DFA over
a compact alphabet, used *only* to answer "does this paragraph contain
any needle at all?". Compliant paragraphs — the dominant case — cost
one table lookup per byte and stop there. The original per-needle loop
still names the matches, so diagnostic order, duplicate handling and
messages come from unchanged code.

Automata are cached by needle list, since `ApplySettings` runs once per
config signature per worker and rebuilding cost more than the matching
it saved. The cache is capped, and the table is declined above 16 MB so
a large repo-supplied word-list falls back to the flat-memory loop
rather than expanding to many times its own size.

A differential test checks the automaton against the `strings.Contains`
oracle over 4,000 randomized needle/text pairs plus the realistic list.

*Guideline:* #data-structures. **Rule benchmark 9219 -> 1079
ns/paragraph (8.5x).**

### 4. Per-file AST collectors regrew from a nil slice

`buildHeadingNodes` and `buildSectionParagraphs` each paid ~log2(n)
regrowth-and-copy allocations to reach n elements, once per file. Both
now take the root's child count as a capacity hint, applied on the
first match so a document that matches nothing still returns `nil` at
zero allocations — the project's "Return `nil`, not `[]T{}`"
convention.

MDS027 drops off `allocBudgetGrandfathered`: with fixes 1, 2 and 4 it
meets the published <= 10 allocs ceiling, so the gate's own
self-removal rule applies, and its per-rule budget moves to 10 with it.

*Guideline:* "Pre-size slices". **-5.4% allocs/op.**

### 5. MDS070's pre-check had no selectivity

`same-file-anchor`'s only gate was "does the source contain a `#`
byte?", which every ATX heading satisfies. So the rule built the full
slug set — a recursive AST descent plus two maps, with a string per
heading — on virtually every file, to resolve fragment links that
almost none contain. An instrumented run measured the set built for
**573 of 577 files to serve 38 fragment links in total**.

It is now built on the first same-file fragment link, with both the AST
and parse-skip walks driving one `anchorChecker` so the two paths
cannot drift.

*Guideline:* "Gate expensive analyzers behind a cheap pre-check".
**-5.2% allocs/op**; a 150-heading document with no fragment link went
from 164 allocations to 6.

## Review

Three rounds of xhigh review ran against the branch; every finding was
addressed. The ones worth calling out:

- **A cache-key collision** that silently disabled MDS056. The key was
  a NUL-joined concatenation on the premise that NUL cannot appear in a
  YAML scalar, which is false — so a needle containing NUL shared a key
  with the two-needle list it splits into. Keys are now
  length-prefixed, with a regression test.
- **A pre-sizing regression** in `ExtractLinks`: a `bytes.Count` of
  `"]("` also matches images and code samples, so a link-free file went
  from zero allocations to one. Reverted there; the astutil collectors
  keep their hints because their bound is tighter.
- **A nil-vs-empty-slice regression** from pre-sizing, and a latent
  `uint8` overflow in the alphabet builder that would have silently
  corrupted the automaton at 256 distinct needle bytes.
- **Unbounded memory** in the matcher cache and the DFA table, both now
  capped — the cache matters for the long-running `mdsmith lsp`.

## Test plan

Re-run in full on the merge commit (`3714fd1`) after main was merged in:

- [x] `go build ./...`, `go vet ./...`
- [x] `go test ./...` — full suite, no regressions
- [x] `go test -race` on the new matcher
- [x] `golangci-lint run ./internal/... ./cmd/...` — 0 issues
- [x] `internal/integration` per-rule allocation budget gate
- [x] `mdsmith check .` — self-lint clean (583 checked, 0 failures)
- [x] All four touched packages at 100% statement coverage
- [x] Red/green test per fix, plus two differential suites (271k cases
      for the target parser, 4k randomized for the matcher)
- [x] `mdsmith check .` output verified byte-identical against the
      current `main`

https://claude.ai/code/session_01EbjcJLfP8k6szThunjBDoZ


jeduden:
🟢 **Merge Queue** — picked up

**Next:** No action needed — you'll get another comment when CI starts on the batch. [View merge queue run](https://github.com/jeduden/mdsmith/actions/runs/33248808435).

jeduden:

Merged into batch branch [`merge-queue/batch-813-1788000774`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-813-1788000774). [View CI run](https://github.com/jeduden/mdsmith/actions/runs/33248838512).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`f104955`](https://github.com/jeduden/mdsmith/commit/f104955bf3719ac23ae4404c4d52715f63fbec02). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/33248838512).

PR #818, opened 2026-08-29
feat(schema): let `filename:` accept a list of globs (OR match)

### LINKED ISSUE #817: schema `<?require filename:?>` cannot express an alternative naming convention
## Context

frit (github.com/jeduden/frit) added a second accepted shape for a
plan file: either a flat `plan/<id>_<slug>.md`, or a folder
`plan/<id>_<slug>/plan.md` holding companion files beside it. Both
shapes get the same `plan` kind and the same schema
(`plan/proto.md`), via `.mdsmith.yml`'s `kind-assignment:` and the
kind's own `path-pattern:` (doublestar, so it can express the union
as `plan/{proto.md,[0-9][0-9]*_*.md,*/plan.md}`).

The schema file itself also carries a `<?require filename: "[0-9]*_*.md"?>`
directive enforcing the same naming convention independently. That
field is decoded into a single Go `string` (`Schema.Filename`) and
matched via stdlib `filepath.Match`, which has no alternation or
brace syntax. Confirmed empirically: passing a YAML sequence —
`filename: ["[0-9]*_*.md", "plan.md"]` — fails to parse:

```
MDS020 invalid schema "plan/proto.md": invalid <?require?> directive:
yaml: unmarshal errors:  line 1: cannot unmarshal !!seq into string
```

So there is no way to express "either this id-prefixed shape, or
exactly this other literal name" through `filename:` alone — the
`path-pattern` field is the only piece of the config that can express
that union today, and it emits a separate diagnostic from `filename:`'s
own MDS020 check.

## Request

Let `filename:` accept either a single glob string (as today, for
backward compatibility) or a list of glob strings, matching if the
basename matches any one of them — the same OR semantics
`unique-frontmatter.include`, `overrides.glob`, `kind-assignment.glob`
and catalog `glob:` already give the rest of `.mdsmith.yml`. Failing
that, CUE-style disjunction syntax (`"[0-9]*_*.md" | "plan.md"`) would
also close the gap, matching how front-matter field constraints are
already expressed elsewhere in a schema.

## Workaround in use

frit's `plan/proto.md` loosens `filename:` to `"*.md"` (effectively a
no-op, since only `.md` files ever reach this check) and relies solely
on the kind's `path-pattern:` for the real naming-shape enforcement.
That works, but it means the schema file's own naming intent is no
longer self-documenting from `filename:` alone.

jeduden:
## What

Closes #817.

A schema's `filename:` constraint could only be a **single** glob, matched by `filepath.Match` (no alternation or brace syntax). There was no way to express "either this id-prefixed shape, or exactly this other literal name" through `filename:` alone — passing a YAML sequence failed to parse:

```
MDS020 invalid schema "plan/proto.md": invalid <?require?> directive:
yaml: unmarshal errors:  line 1: cannot unmarshal !!seq into string
```

`filename:` now accepts **either** a single glob string (unchanged, backward compatible) **or** a list of glob strings, matching when the basename matches any one of them — the same OR semantics `overrides.glob`, `kind-assignment.glob`, and catalog `glob:` already give elsewhere in the config:

```markdown
<?require
filename:
  - "[0-9]*_*.md"
  - "plan.md"
?>
```

or inline on a kind:

```yaml
kinds:
  plan:
    schema:
      filename:
        - "[0-9]*_*.md"
        - "plan.md"
```

## How

The constraint is parsed and validated by **two** code paths, both updated:

- **`internal/schema`** package — inline kind schemas, `extends:` inheritance, and composed multi-source schemas (`Schema.Filename` is now `[]string`).
- **`internal/rules/requiredstructure`** legacy parser — the file-schema path MDS020 uses for a single `proto.md` source without `extends:` (the exact path that produced the issue's error). `schemaConfig.FilenamePattern` is now `FilenamePatterns []string`.

Three shared helpers in `internal/schema` keep decoding and diagnostics consistent across both parsers:

- `DecodeFilenameField` — normalizes a string or YAML sequence to `[]string`.
- `MatchFilename` — OR-matches the basename, reporting the offending glob on a malformed pattern.
- `FilenameExpected` — a single glob keeps the historical `filename matching glob X` wording; several render `filename matching one of globs X, Y`.

Merge semantics: `compose` treats two non-empty lists as conflicting unless they're equal (`slices.Equal`); `extends` lets a child override the parent's list wholesale — both unchanged in spirit from the single-string behavior.

## Tests

- Unit tests for `DecodeFilenameField` / `MatchFilename` / `FilenameExpected` (string, list, empty, malformed, wrong-type).
- List-form parse + validate tests for both parsers, including a match that hits only the *second* glob (proving OR) and the multi-glob "expected" diagnostic.
- New MDS020 good/bad fixtures (`filename-list-match` / `filename-list-mismatch`) exercising the feature end to end.

`go test ./...`, `go vet ./...`, `golangci-lint` (0 issues), the rule allocation-budget test, and `mdsmith check .` all pass.

## Docs

Documented the list form in the enforcing-structure guide (full example), the schemas guide, the section-schema reference, and the MDS020 README settings table.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Qfpat8jP8v3PFnUrM7L3xB)_

### DIFF
--- a/docs/guides/directives/enforcing-structure.md
+++ b/docs/guides/directives/enforcing-structure.md
@@ -82,6 +82,20 @@ basename matching the glob. If they don't, `check`
 reports: `filename "foo.md" does not match required
 pattern "[0-9]*_*.md"`.
 
+`filename:` also accepts a list of globs. The basename
+then passes when it matches any one of them. This is the
+OR match that `overrides.glob` and `kind-assignment.glob`
+already give. Use it when a kind accepts more than one
+naming shape — say an id-prefixed file or a fixed name:
+
+```markdown
+<?require
+filename:
+  - "[0-9]*_*.md"
+  - "plan.md"
+?>
+```
+
 **Schema-only**: `<?require?>` is only recognized in
 schema files. Using it in a normal file emits:
 `<?require?> is only recognized in schema files; this
--- a/docs/guides/schemas.md
+++ b/docs/guides/schemas.md
@@ -75,9 +75,9 @@ the ISO regex; see
 [Schema field types](../reference/schema-types.md)
 for the registered names and how they are matched.
 
-`filename:` is a glob the document basename must
-match. It sits at the top of the schema block (no
-`require:` wrapper).
+`filename:` is a glob — or a list of globs the
+basename may match any of — at the top of the schema
+block (no `require:` wrapper).
 
 `closed: true` makes the scope strict — unlisted
 headings produce a diagnostic. `closed: false` (the
--- a/docs/reference/section-schema.md
+++ b/docs/reference/section-schema.md
@@ -235,8 +235,8 @@ schema:
   closed: <bool>
 ```
 
-- `filename:` — a glob the document basename
-  must match. Top-level; no `require:` wrapper.
+- `filename:` — a glob, or list of globs any of
+  which the basename matches. No `require:` wrapper.
 - `frontmatter:` — per-key CUE constraints.
   Trailing `?` on a key marks it optional.
 - `sections:` — the top-level section list.
--- a/internal/rules/MDS020-required-structure/README.md
+++ b/internal/rules/MDS020-required-structure/README.md
@@ -58,9 +58,9 @@ file's leading comment for the vocabulary.
 Use `<?require?>` in the schema body to declare
 constraints on files validated against this schema:
 
-| Field      | Type   | Description                           |
-| ---------- | ------ | ------------------------------------- |
-| `filename` | string | Glob the document basename must match |
+| Field      | Type           | Description                                       |
+| ---------- | -------------- | ------------------------------------------------- |
+| `filename` | string or list | Glob(s) the document basename must match (any of) |
 
 ```markdown
 <?require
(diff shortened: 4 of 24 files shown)


jeduden:
🟢 **Merge Queue** — picked up

**Next:** No action needed — you'll get another comment when CI starts on the batch. [View merge queue run](https://github.com/jeduden/mdsmith/actions/runs/33248115580).

jeduden:

Merged into batch branch [`merge-queue/batch-818-1787999663`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-818-1787999663). [View CI run](https://github.com/jeduden/mdsmith/actions/runs/33248148493).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`348d95a`](https://github.com/jeduden/mdsmith/commit/348d95a3f5dc3df71f5a0abeb4c90014a795af52). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/33248148493).

PR #819, opened 2026-08-29
fix(security): remove proxy blind spot from MDS072 guarded HTTP client

jeduden:
## What

Closes plan `2608282011` — security hardening batch 2026-08-28.

The MDS072 external-link-check rule's guarded HTTP client previously set
`Proxy: http.ProxyFromEnvironment`. When `HTTP_PROXY` or `HTTPS_PROXY` is set,
Go's transport dials the forward proxy rather than the destination. The
`ssrfControl` dial hook therefore checks the proxy's IP — not the target's. A
forward proxy on a non-restricted IP can then forward the request to an RFC1918
address, cloud-metadata IP (169.254.169.254), or any other address the SSRF
guard was meant to block, without `ssrfControl` ever seeing it.

This fix sets `Proxy: nil` on the guarded transport so every dial is direct and
`ssrfControl` always vets the actual destination.

Preconditions for the vulnerability: MDS072 must be enabled (off by default), a
forward proxy must be configured in the environment, and that proxy must be
willing to forward to restricted ranges. Defense-in-depth, not an exposed
default.

`permissiveClient` (used when `external-allow-internal: true`) is unchanged and
still honors the environment proxy.

## How

- `internal/rules/externallink/probe_net.go`: `Proxy: http.ProxyFromEnvironment`
  → `Proxy: nil` in `buildGuardedClient`, with a doc comment recording the
  invariant so it cannot be silently reintroduced.
- `internal/rules/externallink/probe_net_test.go`: `TestGuardedClientHasNoProxy`
  asserts the transport's `Proxy` field is nil — red against the old code, green
  after the fix.
- `internal/rules/MDS072-external-link-check/README.md`: new "Environment proxy"
  paragraph documenting that the guarded path does not honor the environment proxy
  and why.

Three rounds of xhigh code review were run after implementation; findings about
an always-green supplementary test and misleading comments were addressed by
keeping only the genuine red-then-green structural assertion.

## Test plan

- [x] `TestGuardedClientHasNoProxy` was red against `Proxy: http.ProxyFromEnvironment` and is green with `Proxy: nil`
- [x] `go test ./internal/rules/externallink/...` passes
- [x] `go test ./...` passes (162 packages, all green)
- [x] `mdsmith check .` passes (583 files, 0 failures)

---
_Generated by [Claude Code](https://claude.ai/code/session_01LmgwTtmx5b4VF6aEX783wX)_

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -266,5 +266,5 @@ footer: |
 | 2608091911 | ✅     | haiku  | [Add dedicated unit tests for occurrence and slidevstructure private helpers](plan/2608091911_arch-fix-missing-unit-tests.md)                           |
 | 2608161754 | ✅     | sonnet | [Vendor go-runewidth as a patched fork so post-LUT versions build under tinygo](plan/2608161754_vendor-runewidth-bump-tinygo.md)                        |
 | 2608161914 | 🔲     | haiku  | [Add dedicated unit tests for the 2026-08-16 touched-set tax findings](plan/2608161914_arch-fix-touched-set-unit-tests.md)                              |
-| 2608282011 | 🔲     | sonnet | [Security hardening batch — 2026-08-28](plan/2608282011_security-hardening-batch-2026-08-28.md)                                                         |
+| 2608282011 | ✅     | sonnet | [Security hardening batch — 2026-08-28](plan/2608282011_security-hardening-batch-2026-08-28.md)                                                         |
 <?/catalog?>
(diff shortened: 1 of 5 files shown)


jeduden:
🟢 **Merge Queue** — picked up

**Next:** No action needed — you'll get another comment when CI starts on the batch. [View merge queue run](https://github.com/jeduden/mdsmith/actions/runs/33249337057).

jeduden:

Merged into batch branch [`merge-queue/batch-819-1788001611`](https://github.com/jeduden/mdsmith/tree/merge-queue/batch-819-1788001611). [View CI run](https://github.com/jeduden/mdsmith/actions/runs/33249368998).

jeduden:
✅ **Merge Queue** — merged

This PR landed on `main` via commit [`e2089bd`](https://github.com/jeduden/mdsmith/commit/e2089bdb821c1c55b007e474720ea3ad268b3c6a). [CI run that validated the merge](https://github.com/jeduden/mdsmith/actions/runs/33249368998).

PR #829, opened 2026-08-30
feat: mdsmith move — refactor engine for file moves and rename redesign

jeduden:
Implements the refactor engine redesign ([plan/2607040822](plan/2607040822_refactor-move-rename-redesign.md)): `move` and `rename` become two verbs over one engine that emits a neutral `Plan` — per-file edits keyed by output target plus an optional file operation — which the CLI, LSP server, and WASM engine each execute their own way.

## What's new

- **`mdsmith move <src> <dst>`** relocates a file and rewrites every reference in one step: incoming `[text](path)` links and `[label]: path` ref-def destinations (fragment preserved), the moved file's own outbound relative links (recomputed from the destination directory), and `[[stem]]` wikilinks when the basename changes. A tracked file is staged with `git mv`; otherwise a plain rename. `--dry-run`, text/JSON output, exit codes 0/1/2. Safety: workspace-relative paths only, an existing destination or missing source aborts with nothing written.
- **`mdsmith rename` redesigned** to a single `--as heading|label` (auto-detected when omitted), with `--dry-run`. A path-shaped or missing-symbol request is steered to `mdsmith move`.
- **Engine (`pkg/mdsmith.Session`)** exposes `Rename` and plan-only `Move`, returning a public `RefactorPlan`; both are mirrored into the WASM/JS proxy and reported by `capabilities()`. *(Explicitly requested: the Go API exposes this.)*
- **LSP** advertises `workspace.fileOperations.willRename` / `didRename` filtered to `**/*.{md,markdown}`, answers `willRenameFiles` with a covering `WorkspaceEdit`, and swaps the path in its index on `didRenameFiles`.
- **Obsidian runtime facade** gains `rename` / `move`, tested through the real WASM build.

## Structure

`internal/rename` → `internal/refactor` (`Plan` / `FileOp`, `Move` planner). The index gains `IncomingPathEdges`, an `EdgeWikilink` kind, and `IncomingWikilinkEdges`. The CLI shares one `applyPlan`. `FileOp` execution (git mv / rename) is behind `//go:build !wasm`.

## Tests & validation

`go test ./...`, `golangci-lint`, both WASM budgets, the Obsidian `bun test` suite, and `mdsmith check .` all pass. TDD throughout; each layer has unit + integration coverage (temp-git-repo move tests, three-file LSP `willRenameFiles` test, WASM-backed rename/move).

## Deferred follow-ups

[...]

PR #825, opened 2026-08-29
perf: five hot-path fixes from a high-performance-go audit

jeduden:
## Summary

An audit against `docs/development/high-performance-go.md` turned up several violations of the project's own performance guidelines. Each candidate was verified empirically with `testing.AllocsPerRun` / benchmarks (and, in two cases, ruled *out* after escape analysis showed the Go compiler already stack-allocated the "problem" — see Notes below) before being fixed with red/green TDD. This PR contains the five highest-confidence, safely-scoped fixes:

1. **`internal/rules/catalog` (MDS019, catalog)** — `renderMinimal` built each row as a `+`-concatenated string before a single `WriteString` call. Pre-sizing the `strings.Builder` with `Grow` and writing each part with its own `WriteString` call brings a 50-entry catalog from **58 allocations to 1**, independent of entry count. This repo's own `CLAUDE.md` and `docs/development/index.md` catalogs exercise this exact path.
2. **`internal/rules/noundefinedreferencelabels` (MDS054, default-enabled)** — three identical `fmt.Sprintf("... %q ...", string(label))` call sites deduplicated into one `undefinedRefMessage` helper built with `strconv.Quote` + concatenation: **3 allocs/317ns → 1 alloc/168ns** per unresolved reference label (e.g. after a bulk rename).
3. **`internal/rules/samefileanchor` (MDS070, default-enabled)** — the heading-slug disambiguation loop built each `"slug-N"` collision candidate via `slug + "-" + strconv.Itoa(n)`; building it into a reused `[]byte` with `strconv.AppendInt` measured **~410ns → ~355ns** for 5 collision attempts (allocation count unchanged — the candidate must become a string once per attempt regardless of construction method).
4. **`internal/rules/tocdirective` (MDS035)** — `buildMessage`'s single-substitution `fmt.Sprintf` replaced with plain concatenation: **~139ns → ~64ns**, same allocation count.
5. **`internal/rules/notrailingpunctuation` (MDS017, default-enabled)** — the same `fmt.Sprintf(%q, ...)` pattern as #2, fixed the same way: **~206ns/3 allocs → ~125ns/2 allocs**.

Every fix is backed by a permanent test: either a new `TestXxx`/`BenchmarkXxx` in the package's existing `alloc_test.go` convention, or a message-format-parity test guarding the wording didn't change.

### Notes: two candidates ruled out

Two other candidates surfaced by the initial scan turned out to be false leads once measured, and are **not** included:

[...]

PR #838, opened 2026-09-05
perf: fix top 5 high-performance-go violations (move/rename engine + MDS059 gate)

jeduden:
## Summary

An audit against `docs/development/high-performance-go.md` found that most of the codebase has already been through prior optimization rounds (PRs #811, #813, #825, #828, #833, #835). This round focused parallel scans on the areas those rounds didn't touch: the newest subsystems (the `mdsmith move`/rename engine and its LSP callers, PR #829/#827) and the remaining unaudited rule packages. Five issues were selected and fixed, each with a red/green test that pins the regression:

1. **`internal/rules/blockquotewhitespace`** (MDS059, default-enabled) — `checkBlankBetween` walked the full AST (or Layer 0 spans) on every workspace file to look for a blank line between sibling blockquotes, even on files with zero `>` marker lines, which can never trigger the violation. The MD027 half of `Check` already scans every line for a leading `>`; the fix reuses that scan as a `sawBlockquote` gate instead of re-deriving it. **7140ns → 689ns** on a 200-section no-blockquote fixture.

2. **`internal/index`** — `BacklinksFor` and `sortEdgesBySource` each drove `sort.Slice` (`reflect.Swapper` internally) — new call sites added for the move engine that were never migrated to the `slices.SortFunc` pattern already used elsewhere (`internal/backlinks`, `internal/fix`). Consolidated into one `slices.SortFunc`-based `sortEdgesBySource`, removing the duplicate comparator too. **3 allocs/op → 0**.

3. **`internal/refactor` + `internal/lsp`** — `stableSortEdits` (heading rename) and `sortTextEditsBottomUp` (LSP rename/`willRenameFiles`) each drove `sort.SliceStable`, the same reflect-in-hot-path anti-pattern, on the move/rename engine's edit-ordering path. Replaced with `slices.SortStableFunc`. **3 allocs/op → 0** in each package.

4. **`internal/lsp`** — `resolveURIAndSource` resolved an open buffer via `openURIs()` (one lock + a slice allocation covering every open document) followed by `docs.get(uri)` per candidate (a second lock plus a struct-copy allocation per document, even on a miss) — O(open-docs) lock acquisitions and allocations on every call, which the move engine calls once per workspace file it touches. Added `documentStore.findByPath`, which scans under a single read lock and copies only the document that actually matches. **51 allocs/op (50 open docs) → 0**.

Each fix follows red/green TDD: a failing test (or ns/op-budget test verified red by temporarily reverting the fix) precedes the change, per `docs/development/high-performance-go.md`'s benchmark→profile→fix loop and the "reflect in hot paths" / "gate expensive analyzers behind a cheap pre-check" / "skip work you don't need" patterns it documents.

### Deferred (out of scope for this PR)

The scan also surfaced a larger issue in the same move engine: a batch file rename (`internal/lsp/fileops.go`'s `handleWillRenameFiles`) calls `refactor.Move` once per renamed file, and each call's `appendRefDefPathEdits` re-scans and re-parses every workspace file — O(N×M) for N renamed files across an M-file workspace. Fixing it safely requires a batch-aware API (sharing the per-file parse across the whole rename batch) rather than a single-function patch, so it's left for a dedicated follow-up rather than risked here.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go run ./cmd/mdsmith check .` (588 files, 0 failures)
- [x] Each fix verified red (test fails without the change) → green (passes with it)
- [x] `-race` on changed packages: intermittent, pre-existing suite flakiness under `-race` (goroutine-scheduling-dependent, not deterministic) reproduces byte-for-byte on an unmodified `origin/main` worktree with no diff applied — none of it is attributable to this change. Repeated isolated runs of the affected packages on baseline pass clean; the failures only surface under some full-package `-race` runs regardless of this diff.

https://claude.ai/code/session_014Tt5GmpywaKpwm7YhA2WvM

### DIFF
--- a/internal/index/index.go
+++ b/internal/index/index.go
@@ -15,9 +15,10 @@
 package index
 
 import (
+	"cmp"
 	"path/filepath"
 	"runtime"
-	"sort"
+	"slices"
 	"strings"
 	"sync"
 )
@@ -467,15 +468,7 @@ func (i *Index) BacklinksFor(file string) []Edge {
 		return nil
 	}
 	edges := i.IncomingEdges(file, "")
-	sort.Slice(edges, func(a, b int) bool {
-		if edges[a].SourceFile != edges[b].SourceFile {
-			return edges[a].SourceFile < edges[b].SourceFile
-		}
-		if edges[a].SourceLine != edges[b].SourceLine {
-			return edges[a].SourceLine < edges[b].SourceLine
-		}
-		return edges[a].SourceCol < edges[b].SourceCol
-	})
+	sortEdgesBySource(edges)
 	return edges
 }
 
@@ -554,15 +547,16 @@ func (i *Index) IncomingWikilinkEdges(stem string) []Edge {
 
 // sortEdgesBySource orders edges by (SourceFile, SourceLine, SourceCol)
 // so move and backlink queries return a stable, reviewable order.
+// slices.SortFunc compares the concrete Edge values directly, unlike
+// sort.Slice, which drives reflect.Swapper internally (see
+// docs/development/high-performance-go.md, "reflect in hot paths").
 func sortEdgesBySource(edges []Edge) {
-	sort.Slice(edges, func(a, b int) bool {
-		if edges[a].SourceFile != edges[b].SourceFile {
-			return edges[a].SourceFile < edges[b].SourceFile
-		}
-		if edges[a].SourceLine != edges[b].SourceLine {
-			return edges[a].SourceLine < edges[b].SourceLine
-		}
-		return edges[a].SourceCol < edges[b].SourceCol
+	slices.SortFunc(edges, func(a, b Edge) int {
+		return cmp.Or(
+			cmp.Compare(a.SourceFile, b.SourceFile),
+			cmp.Compare(a.SourceLine, b.SourceLine),
+			cmp.Compare(a.SourceCol, b.SourceCol),
+		)
 	})
 }
 
--- a/internal/index/race_off_test.go
+++ b/internal/index/race_off_test.go
@@ -1,0 +1,5 @@
+//go:build !race
+
+package index
+
+const raceEnabled = false
--- a/internal/index/race_on_test.go
+++ b/internal/index/race_on_test.go
@@ -1,0 +1,5 @@
+//go:build race
+
+package index
+
+const raceEnabled = true
--- a/internal/index/sortnoreflect_test.go
+++ b/internal/index/sortnoreflect_test.go
@@ -1,0 +1,38 @@
+package index
+
+import "testing"
+
+// TestSortEdgesBySource_NoReflectSort pins the allocation cost of
+// sortEdgesBySource, which drove sort.Slice — reflect.Swapper
+// internally — the "reflect in hot paths" anti-pattern in
+// docs/development/high-performance-go.md, already fixed the same way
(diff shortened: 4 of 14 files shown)


jeduden:
## Security review — no findings

Ran the `mdsmith-security-review` methodology (PR/diff mode) over this diff. **No security findings.**

Behavior-preserving hot-path refactors, none crossing a trust boundary:

- **`internal/index/index.go`** — `sortEdgesBySource` `sort.Slice` → `slices.SortFunc`; identical (SourceFile, SourceLine, SourceCol) ordering.
- **`internal/lsp/documents.go` + `rename.go`** — new `findByPath` replaces the `openURIs()`+`get()` scan. Reviewed the concurrency: it holds `mu.RLock` for the whole scan and copies only the matching document, and the `match` callback is a pure path comparison that takes no locks, so there's no new race or lock-ordering issue. The returned copy shares the underlying `text` slice exactly as `get()` already did — no new aliasing.
- **`internal/refactor/heading.go`, `internal/lsp/rename.go`** — `sort.SliceStable` → `slices.SortStableFunc`; same bottom-up edit ordering that keeps splice offsets valid.
- **`blockquotewhitespace/rule.go` (MDS059)** — adds a `sawBlockquote` early-return gate before the MD028 AST walk. Confirmed it is conservative: `sawBlockquote` is set for any line whose first non-blank byte is `>` *before* the code-fence check, so a `>` inside a code fence still disables the skip. Since an MD028 sibling-blockquote violation always requires a `>`-prefixed line, the gate can never skip a document that could produce the diagnostic — no lost diagnostics.

No new `regexp`, no path/traversal changes, no recipe execution, no network, no unbounded allocation. Sign-off: safe to merge from a security standpoint.

---

jeduden:
## Security review — no findings

Scope: PR diff only (`perf: move/rename engine + MDS059 gate`), reviewed with the mdsmith-security-review methodology (threat model §0 baseline defenses + surfaces touched).

Surfaces mapped: LSP server, CLI/rule engine. Directive engine, distribution, and Git integration are untouched by this diff.

Traced items:

- **`internal/lsp/documents.go` `findByPath` + `resolveURIAndSource`** — read-only lookup. It replaces the prior `openURIs()`+`get()` two-step with a single `RLock`ed scan using the same normalized-workspace-relative-path predicate. `openURIs()` already iterated the map in unspecified order, so this introduces no order dependence, no new file-write or exec sink, and it is actually safer against a concurrent `didClose` (no window between enumerate and fetch). Only one open buffer can match a given normalized `rel`.
- **Sort refactors** (`sortEdgesBySource`, `stableSortEdits`, `sortTextEditsBottomUp`) — `sort.Slice`→`slices.SortFunc/SortStableFunc` with identical comparators. The rename/move edit offsets and bounds math are unchanged, so no path-math regression on the in-place-rewrite path the threat model flags.
- **`blockquotewhitespace` MD028 gate** — skips the AST/Layer-0 walk when no `>`-prefixed marker line was seen. MD028 requires two sibling blockquotes, which cannot exist without marker lines, so no violation is missed. Effect is DoS-reducing (less work on hostile input), not DoS-introducing.

None of the §0 baseline defenses (recipe non-execution, document-content→command isolation, build-output validation, symlink default-deny, Workspace Trust, no postinstall) is affected. No security issues found; safe to merge on this axis.

_Confidence: confirmed (code paths traced against the local tree)._

---

